PR #16, opened 2021-11-30
Feat: Keep major and minor versions synced with the appropriate commit

ja-bravo:
This PR adds a new input named that when enabled, updates the major and minor tags in GitHub to the latest appropriate commit as mentioned in https://github.com/fastify/github-action-merge-dependabot/issues/112#issue-1065370381

Example:

1. Version v3.0.2 is released
2. This feature updates the tag v3 to point to the same commit as v3.0.2

Tasks

- [x] Add a new input to control if this feature is enabled
- [x] Handle major versions
- [x] Handle minor versions
- [x] Update PR template to inform the user if this feature is enabled
- [x] Update documentation
- [x] Add automated tests

Closes[ fastify/github-action-merge-dependabot/112](https://github.com/fastify/github-action-merge-dependabot/issues/112)


release.js
```diff
@@ -55,12 +55,34 @@ module.exports = async function ({ github, context, inputs, callApi }) {
...
+      method: 'PATCH',
+      body: {
+        version: version,
+        releaseId: id,
+      },
+    })
+
+    const syncInput = inputs['sync-major']
+    const syncMajor =
+      syncInput === 'true' || syncInput === 'True' || syncInput === 'TRUE'
```
simoneb, line 70:
```suggestion
    const syncMajor = /true/i.test(inputs['sync-major'])
```

release.js
```diff
@@ -55,12 +55,34 @@ module.exports = async function ({ github, context, inputs, callApi }) {
...
+      await run('git', ['push', 'origin', `:refs/tags/${major}`])
+      await run('git', [
+        'tag',
+        '-fa',
+        `"${major}"`,
+        '-m',
+        `"Update tag ${major}"`,
+      ])
+      await run('git', ['push', 'origin', `--tags`])
+    }
```
simoneb, line 84: if you want to avoid manual parsing consider using the semver npm package which is a more reliable alternative to manually parsing and interpreting version numbers unless you're only doing very basic manipulation

release.js
```diff
@@ -55,12 +55,34 @@ module.exports = async function ({ github, context, inputs, callApi }) {
...
+  } catch (err) {
+    console.error('Unable to publish the release', err.message)
```
simoneb, line 86: note that if the action can't do what it's trying to do it's fine to error. catching this will not error. also note that there are specific helpers for logging and handling errors in github actions inside the actions/core package. See https://github.com/actions/toolkit

release.js
salmanm: I second using `semver` too. You could do

```js
const { major, minor } = require('semver').parse(version)
```

action.yml
```diff
@@ -36,6 +36,10 @@ inputs:
     description: "Url of the API where the application is running"
     required: false
     default: "https://optic-release-automation-ocrlhra4va-ue.a.run.app/"
+  sync-major:
```
salmanm, line 39: Can we also update the minor in case of a patch version?

ja-bravo: Yeah, on it. I'll change the name of the input to reflect what it does better

salmanm:
Aye, I'm late to the party, new changes cover all my inputs.

release.js
```diff
@@ -55,12 +57,31 @@ module.exports = async function ({ github, context, inputs, callApi }) {
...
+        const parsed = semver.parse(version)
+        const major = parsed.major
+        const minor = parsed.minor
+
+        if (major !== 0) await tagVersionInGit(`v${major}`)
+        if (minor !== 0) await tagVersionInGit(`v${major}.${minor}`)
+      }
+    } catch (err) {
+      core.setFailed(`Unable to update the semver tags ${err.message}`)
+    }
```
ja-bravo, line 83: Wondering if this should happen regardless of if publishing the release failed. If the maintainer sets the draft release as published manually after this action runs, the major and minors tags would already be updated.

release.js
```diff
@@ -55,12 +57,31 @@ module.exports = async function ({ github, context, inputs, callApi }) {
...
+      },
+    })
+
+    try {
+      const syncVersions = /true/i.test(inputs['sync-semver-tags'])
+
+      if (syncVersions) {
+        const parsed = semver.parse(version)
+        const major = parsed.major
+        const minor = parsed.minor
```
simoneb, line 76: prefer destructuring

util.js
```diff
@@ -32,3 +32,19 @@ exports.runSpawn = function runSpawn({ cwd } = {}) {
     })
   }
 }
+
+exports.runSpawn = runSpawn
+
+exports.tagVersionInGit = async function (version) {
+  const run = runSpawn()
+
+  await run('git', ['push', 'origin', `:refs/tags/${version}`])
```
simoneb, line 41: what does this do? I think publishing a release already creates a tag. let's make sure this doesn't screw anything up

util.js
```diff
@@ -32,3 +32,19 @@ exports.runSpawn = function runSpawn({ cwd } = {}) {
...
+  await run('git', [
+    'tag',
+    '-fa',
+    `"${version}"`,
+    '-m',
+    `"Update tag ${version}"`,
```
simoneb, line 47: it would be useful if you could explain why you've chosen to provide these flags, it is not obvious otherwise.

my doubts are:

- why do we need -a?
- why do we need -m?

bump.js
```diff
@@ -11,18 +12,28 @@ const actionPath = process.env.GITHUB_ACTION_PATH
 const tpl = fs.readFileSync(path.join(actionPath, 'pr.tpl'), 'utf8')
 
 const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
+  const tagsToBeUpdated = []
+  const parsed = semver.parse(newVersion)
+  const major = parsed.major
+  const minor = parsed.minor
```
simoneb, line 18: prefer desctructuring

release.js
simoneb: no right or wrong here, but I do agree that it may make sense to do the tagging anyway. then let's move the tagging before the release creation.

let's also avoid nested try/catch please

simoneb (commented):
See inline comments. 

As part of this PR, can we please introduce automated tests?

Let's use tap and inspire from how we're doing it in https://github.com/fastify/github-action-merge-dependabot.

We have no automated testing now but before we roll this out to the wild we need to have it in place.

util.js
salmanm: If we had an old `v2` tag, we can't force update it I think. So this line deletes it from the remote and then next line adds it again pointing to the current commit.

ja-bravo: That's the idea yeah. Running the same code without that line will give you the following error
```
 ! [rejected]        v5 -> v5 (already exists)
error: failed to push some refs to 'https://github.com/ja-bravo/optic-action-test.git'
hint: Updates were rejected because the tag already exists in the remote
```

util.js
ja-bravo: The flags -a and -m are needed to make it an annotated tag with a message. 
They're not needed if we want to leave the default tag message ([OPTIC-RELEASE-AUTOMATION] release/vX.X.X)

ja-bravo:
There's still some more work to do on the tests, I'll get that done asap.

bump.js
```diff
@@ -3,30 +3,40 @@
...
+  if (major !== 0) tagsToBeUpdated.push(`v${major}`)
+  if (minor !== 0) tagsToBeUpdated.push(`v${major}.${minor}`)
+
   // Should strictly contain only non-sensitive data
   const releaseMeta = {
     id: draftRelease.id,
     version: newVersion,
     npmTag: inputs['npm-tag'],
     opticUrl: inputs['optic-url'],
+    tagsToUpdate: tagsToBeUpdated.join(', '),
```
simoneb, line 28: Why are we storing these in `releaseMeta`? This is an object that's used to carry information across between the time the PR is created to when it's merged, but it doesn't look like we're using this, because when we're releasing we're using the semver again to extract the tags we need to update. Hence, we have 2 options:

- either we don't store them in releaseMeta and we simply provide them as an input to `template()`
- we store them in releaseMeta and we populate them conditionally based on whether inputs['sync-semver-tags'] is true or not, but then in the release we read them from releaseMeta instead of computing them again

ja-bravo: Good catch, I've gone with the first option

simoneb (approved):

test/log.test.js
```diff
@@ -0,0 +1,53 @@
...
+tap.test('logDebug calls @actions/core/debug', async t => {
+  const { logger, coreStub } = setup()
+  logger.logDebug('Debug')
+  t.ok(coreStub.debug.calledOnce)
+  t.ok(coreStub.error.notCalled)
+})
+
+tap.test('logError calls @actions/core/error', async t => {
+  const { logger, coreStub } = setup()
+  logger.logError('Debug')
```
Eomm, line 36:
```suggestion
  logger.logError(new Error('not a string'))
```

test/release.test.js
```diff
@@ -0,0 +1,360 @@
...
+      ...DEFAULT_ACTION_DATA,
+      context: {
+        ...DEFAULT_ACTION_DATA.context,
+        payload: {
+          ...DEFAULT_ACTION_DATA.context.payload,
+          pull_request: {
+            ...DEFAULT_ACTION_DATA.context.payload.pull_request,
+            user: { login: 'not_the_correct_one' },
+          },
+        },
```
Eomm, line 233: Do you think this could be more clear for the reader?

```
const payload = clone(DEFAULT_ACTION_DATA)
payload.context.payload.pull_request.user = { login: 'not_the_correct_one' }
```

Eomm (commented):
LGTM

minor suggestions

test/release.test.js
ja-bravo: You're absolutely right. I've updated all big nested objects to use a cloning func

simoneb:
We can merge. Because the issue lives in another repo it won't be closed automatically, let's not forget to close it manually

simoneb:
I will do it, thanks.

PR #24, opened 2021-12-05
Multiple tokens

### LINKED ISSUE #18: Support multiple npm/optic tokens
We want to be able to allow multiple people to release npm packages. This means we need multiple NPM tokens and multiple Optic tokens (if needed).

- To do this, the action should support both of the above either as single tokens (as it does now) or multiple.
- Because action inputs cannot be arrays, we need to decide on an encoding format for arrays of values. Because those values are coming from repo secrets, it should also be able to store them there. I would suggest using comma separated values, which should be safe as neither npm or optic tokens should contain commas

Once we have a way to provide arrays of tokens, the action must:

- parse them correctly
- [zip](https://en.wikipedia.org/wiki/Zipping_(computer_science)) the two arrays and validate that they have the same cardinality
- when doing the npm publish, iterate through the pairs and implement the same behavior as [Promise.race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race)

sameer-coder:
This PR adds support for multiple npm/optic tokens.

**Usage:**
- You need to define npm and optic tokens in the Github secrets for each user that will receive the otp
- The `name` of the npm/optic secret for each user should start with `NPM_TOKEN_` and `OPTIC_TOKEN_`
- The `value` of the npm/optic secret for each user should be of the format `<github-username>:<token>`
- Default tokens can be provided with `NPM_TOKEN` and `OPTIC_TOKEN` secrets which will be used if individual user token secrets are not defined (backwards compatibility).

**How does it work?**
The action reads all secrets and checks if the value of a secret is starts with the `github.actor`. If it does then it uses that value for npm/optic token. If there are no matches and `NPM_TOKEN`/`OPTIC_TOKEN` secret is specified then this value is used instead.

Implementation is open for discussion.

Closes #18


README.md
```diff
@@ -41,6 +41,7 @@ jobs:
     steps:
       - uses: nearform/optic-release-automation-action@main # you can use a tag instead of main
         with:
+          all-secrets: ${{ toJson(secrets) }}
```
simoneb, line 44: I'm not 100% sure that this is a viable option. This means that somebody using the action will have to trust sending all the secrets to the action. Before deciding that we don't have other options, have we considered alternatives?

sameer-coder: I suspected this might be an issue even though the action is public. I can try to filter the secrets using regex(there's an action for that) and only send the NPM/Optic token related secrets to the action.

simoneb: Yes, let's see if there's anything else here which could help https://docs.github.com/en/actions/learn-github-actions/expressions

README.md
```diff
@@ -39,8 +39,25 @@ jobs:
...
+      - name: Run action
+        run: |
+          NPM_TOKEN=$(echo '${{ steps.gettokens.outputs.tokens }}' | jq -R '. | fromjson | to_entries[] | select((.key | startswith("NPM_TOKEN_")) and (.value | startswith("${{github.actor}}"))) | .value | split(":")[1]')
+          OPTIC_TOKEN=$(echo '${{ steps.gettokens.outputs.tokens }}' | jq -R '. | fromjson | to_entries[] | select((.key | startswith("OPTIC_TOKEN_")) and (.value | startswith("${{github.actor}}"))) | .value | split(":")[1]')
+
+          echo ::add-mask::$NPM_TOKEN
+          echo ::add-mask::$OPTIC_TOKEN
+
+          echo USER_NPM_TOKEN=$(echo $NPM_TOKEN) >> $GITHUB_ENV
+          echo USER_OPTIC_TOKEN=$(echo $OPTIC_TOKEN) >> $GITHUB_ENV
```
simoneb, line 56: thanks for looking into ways to accomplish this. unfortunately, I don't think that asking users to include this code when using the action is a viable option, we need to come up with an alternative approach to support multiple tokens.

Eomm:
I would propose an option.

As dependabot has its own `dependabod.yml`, we could define a `release-automation.yml` configuration file that we can load at startup. in that file, the user can list the users and the mapping between secrets' name and env var names so the dev is the director here too.

In this way, we access only what we need.

When maintainer-A run it, the action is still reading the maintainer-B tokens: I'm thinking if this could be optimized even more.

Example:

```
jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: nearform/optic-release-automation-action@main
        with:
          config: './github/release-automation.yml'
          github-token: ${{secrets.github_token}}
          npm-tag: ${{ github.event.inputs.tag }}
          env:
            KEY1: ${{ secrets.NPM_TOKEN_USER_FOO }}
            KEY2: ${{ secrets.OPTIC_TOKEN_FOO }}
            KEY_whatever: ${{ secrets.NPM_BAR }}
            KEY_hurra: ${{ secrets.OPTIC_TOKEN_BAR }}
```

And then the `release-automation.yml` file

```yml
users:
 - name: 'maintainer1'
   npm_token: 'KEY1'
   optic_token: 'KEY2'
 - name: 'maintainer-two'
   npm_token: 'KEY_whatever'
   optic_token: 'KEY_hurra'
```

sameer-coder:
> I would propose an option.
> 
> As dependabot has its own `dependabod.yml`, we could define a `release-automation.yml` configuration file that we can load at startup. in that file, the user can list the users and the mapping between secrets' name and env var names so the dev is the director here too.
> 
> In this way, we access only what we need.
> 
> When maintainer-A run it, the action is still reading the maintainer-B tokens: I'm thinking if this could be optimized even more.
> 
> Example:
> 
> ```
> jobs:
>   release:
>     runs-on: ubuntu-latest
>     steps:
>       - uses: nearform/optic-release-automation-action@main
>         with:
>           config: './github/release-automation.yml'
>           github-token: ${{secrets.github_token}}
>           semver: ${{ github.event.inputs.semver }}
>           npm-tag: ${{ github.event.inputs.tag }}
>           env:
>             KEY1: ${{ secrets.NPM_TOKEN_USER_FOO }}
>             KEY2: ${{ secrets.OPTIC_TOKEN_FOO }}
>             KEY_whatever: ${{ secrets.NPM_BAR }}
>             KEY_hurra: ${{ secrets.OPTIC_TOKEN_BAR }}
> ```
> 
> And then the `release-automation.yml` file
> 
> ```yaml
> users:
>  - name: 'maintainer1'
>    npm_token: 'KEY1'
>    optic_token: 'KEY2'
>  - name: 'maintainer-two'
>    npm_token: 'KEY_whatever'
>    optic_token: 'KEY_hurra'
> ```

Sounds good to me. @simoneb what do you think?

simoneb:
Thanks @Eomm @sameer-coder for the work and thinking on this. I think nonetheless that we're still not there, the proposals we have on the table so far all require considerable work in setting up this action from a user's perspective and we need to iterate on this proposals to come up with something better, which I am fairly confident we can do if we take a little more time to think on it.

sameer-coder:
I can think of 2 other solutions which might make it easier for the end user.

- Enfore a naming convention for user npm/optic token. Example: `NPM_TOKEN_<github_user_id>`. Since GitHub user id is available in the action and is numeric it will be a valid secret name and should be easy to implement. Not sure if users will like the extra step of getting the user-id before saving the secret
- Maybe we can create a new action called `filter-secrets` which takes a GitHub username and secrets as input and returns the matching secret which is in our case is npm/optic token

@Eomm @simoneb Do you think any of these are feasible solutions?

simoneb:
We need to do a bit of brainstorming on this. I'll schedule something

simoneb (commented):
I like this new version! There's only one thing I would change. I would not introduce two new inputs, can we keep only npm-token and optic-token and delegate to the user to decide how to use them without imposing them to use a user-specific token and a generic one?

we could provide a few examples of usage:

- only default tokens, e.g. `npm-token: ${{secrets.NPM_TOKEN}}`
- only user-related tokens: e.g. `npm-token: ${{ secrets[format('NPM_TOKEN_{0}', github.actor)] }}` (note, no need to use env)
- both (please check this would work: `npm-token: ${{ secrets[format('NPM_TOKEN_{0}', github.actor)] ||  secrets.NPM_TOKEN }}`

action.js
```diff
@@ -8,20 +8,23 @@ const { logError } = require('./log')
 module.exports = async function ({ github, context, inputs }) {
   const run = runSpawn()
 
-  if (inputs['npm-token']) {
+  const npmToken = inputs['user-npm-token'] || inputs['npm-token']
+  const opticToken = inputs['user-optic-token'] || inputs['optic-token']
```
Eomm, line 12: Are we still using the `user-optic-token`?

action.yml
```diff
@@ -13,8 +13,8 @@ inputs:
     description: "URL if you have a custom application that serves OTP"
     required: false
     default: "https://optic-zf3votdk5a-ew.a.run.app/api/generate/"
-  optic-token:
-    description: "Your Optic token. You can add your Npm secret to the Optic app, generate a new token and pass it to this input. Required only if you want to pubish to Npm and have MFA enabled for your account"
+  user-optic-token:
```
Eomm, line 16: This should be reverted

README.md
```diff
@@ -56,6 +56,15 @@ The above workflow (when manually triggered) will
 ![image](https://user-images.githubusercontent.com/2510597/140506212-4938e44d-0662-4dc5-9fb1-c3f59fe075a6.png)
 
 - When you merge this PR, it will request an Npm OTP from Optic. (If you close the PR, nothing will happen)
+- You need to define Npm and Optic tokens in the Github secrets for each user that will receive the otp
+- You can either provide npm and optic tokens of the user who has initiated the workflow or provide default tokens or both.  
+**Some examples**:  
+  - Use only default tokens:   
+    *e.g.* npm-token: ${{secrets.NPM_TOKEN}}
```
Eomm, line 63:
```suggestion
    *e.g.* npm-token: ${{ secrets.NPM_TOKEN }}
```

README.md
```diff
@@ -56,6 +56,15 @@ The above workflow (when manually triggered) will
 ![image](https://user-images.githubusercontent.com/2510597/140506212-4938e44d-0662-4dc5-9fb1-c3f59fe075a6.png)
 
 - When you merge this PR, it will request an Npm OTP from Optic. (If you close the PR, nothing will happen)
+- You need to define Npm and Optic tokens in the Github secrets for each user that will receive the otp
+- You can either provide npm and optic tokens of the user who has initiated the workflow or provide default tokens or both.
```
Eomm, line 60: What do you think moving this into the `example` section highlighting the "multiple user" feature.

I'm afraid here the user could miss it

sameer-coder: I think this is already in the example section. Are you suggesting we show separate(single and multi user) workflows in the example code?

sameer-coder: Or have a seperate multi-user section where we document this?

Eomm: Yes, this would give to this feature more attention I think

Eomm (approved):

README.md
```diff
@@ -56,8 +56,53 @@ The above workflow (when manually triggered) will
 ![image](https://user-images.githubusercontent.com/2510597/140506212-4938e44d-0662-4dc5-9fb1-c3f59fe075a6.png)
 
 - When you merge this PR, it will request an Npm OTP from Optic. (If you close the PR, nothing will happen)
-- Upon successful retrieval of the OTP, it will publish the package to Npm.
+- You need to define Npm and Optic tokens in the Github secrets for each user that will receive the otp
```
simoneb, line 59: note that this is not mandatory, because if you don't specify those inputs simply there won't be any publish to npm, which is what we want because this action can be used to release other things (e.g. applications or github actions)

README.md
```diff
@@ -56,8 +56,53 @@ The above workflow (when manually triggered) will
...
 - Create a Github release with change logs (You can customize release notes using [release.yml](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration))
+- Upon successful retrieval of the OTP, it will publish the package to Npm.
+
+### Multiple user scenario
+
+In case there are multiple users who have access to trigger the release automation action, you can define npm and optic tokens for different
```
simoneb, line 65: let's be consistent with casing when mentioning npm and optic. let's use the same casing everywhere

README.md
```diff
@@ -56,8 +56,53 @@ The above workflow (when manually triggered) will
...
+users in GitHub secrets. The naming convention to be used when creating the secrets is `NPM_TOKEN_<github-username>` and 
+`OPTIC_TOKEN_<github-username>`. In the workflow you can provide npm and optic tokens of the user who has initiated the workflow or provide default tokens or both.  
+
+**Some examples**:
```
simoneb, line 69: use `####` header and don't use "some examples" since the example is only one

README.md
```diff
@@ -56,8 +56,53 @@ The above workflow (when manually triggered) will
...
+      semver:
+        description: "The semver to use"
+        required: true
+        default: "patch"
+      tag:
+        description: "The npm tag"
+        required: false
+        default: "latest"
+  pull_request:
+    types: [closed]
```
simoneb, line 92: you can omit all this part since it's the same as the default example

simoneb (changes requested):

simoneb (changes requested):
There is no mention of the limitations of secret names for usernames which are not valid secret names

README.md
```diff
@@ -56,8 +56,38 @@ The above workflow (when manually triggered) will
 ![image](https://user-images.githubusercontent.com/2510597/140506212-4938e44d-0662-4dc5-9fb1-c3f59fe075a6.png)
 
 - When you merge this PR, it will request an Npm OTP from Optic. (If you close the PR, nothing will happen)
-- Upon successful retrieval of the OTP, it will publish the package to Npm.
+- (Optional) You can define Npm and Optic tokens in Github secrets for each user that will receive the otp. This is required only in case you want to publish to Npm.
```
simoneb, line 59: Please always use correct casing! (github and otp which is an acronym)

```suggestion
- (Optional) You can define Npm and Optic tokens in GitHub secrets for each user that will receive the OTP. This is required only in case you want to publish to Npm.
```

README.md
```diff
@@ -56,8 +56,38 @@ The above workflow (when manually triggered) will
...
 - Create a Github release with change logs (You can customize release notes using [release.yml](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration))
+- Upon successful retrieval of the OTP, it will publish the package to Npm.
+
+### Multiple user scenario
+
+In case there are multiple users who have access to trigger the release automation action, you can define Npm and Optic tokens for different users in GitHub secrets. The naming convention to be used when creating the secrets is `NPM_TOKEN_<github-username>` and
```
simoneb, line 65: This is not entirely accurate. We don't enforce this naming convention in any way, it's just an option. Users could name their secrets whatever they want really! Here's we're providing an example of a way to use different tokens depending on the user who merged the pull request

README.md
simoneb: @sameer-coder please address this comment (simply accept the suggestion)

README.md
```diff
@@ -56,8 +56,38 @@ The above workflow (when manually triggered) will
...
 - Create a Github release with change logs (You can customize release notes using [release.yml](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration))
+- Upon successful retrieval of the OTP, it will publish the package to Npm.
+
+### Multiple user scenario
+
+In case there are multiple users who have access to trigger the release automation action, you can define Npm and Optic tokens for different users in GitHub secrets. The naming convention to be used when creating the secrets is `NPM_TOKEN_<github-username>` and 
+`OPTIC_TOKEN_<github-username>`. In the workflow you can provide Npm and Optic tokens of the user who has initiated the workflow or provide default tokens or both.    
+
+#### Example:  
+  - Use only default tokens:
```
simoneb, line 69: the examples in this section must be formatted as code

simoneb (changes requested):
See inline comments

README.md
sameer-coder: > @sameer-coder please address this comment (simply accept the suggestion)

I have formatted the examples as code. Is there anything else you feel should be changed here?

README.md
sameer-coder: You are right! The secret name can really be anything!
I have updated the docs

README.md
```diff
@@ -56,14 +56,43 @@ The above workflow (when manually triggered) will
...
+      - uses: nearform/optic-release-automation-action@main # you can use a tag instead of main
+        with:
+          github-token: ${{ secrets.github_token }}
+          npm-token: ${{ secrets[format('NPM_TOKEN_{0}', github.actor)] || secrets.NPM_TOKEN }}
+          optic-token: ${{ secrets[format('OPTIC_TOKEN_{0}', github.actor)] || secrets.OPTIC_TOKEN }}
+          semver: ${{ github.event.inputs.semver }}
+          npm-tag: ${{ github.event.inputs.tag }}
+```
+
+*Note*: The multiple user feature does not work if the GitHub username contains hyphens(-) because GitHub secrets do not allow hypens in the secret's name.
```
simoneb, line 89:
```suggestion
*Note*: Not all symbols that can be used in GitHub usernames are valid in secret names. One such example is the hyphen symbol (`-`). In such cases, this approach will not work.  
```

simoneb (approved):
I added one suggestion about the wording to use when mentioning the invalid symbols in secret names. Please include it before merging.

PR #29, opened 2021-12-09
Delete release branch

### LINKED ISSUE #23: Release branch should be deleted when the PR is closed (without merging)
If the branch is not deleted, the next trigger of the action will error as it will try to recreate the branch

ja-bravo:
This PR changes the action to delete the release/VX.X.X branch if the PR is closed.

Closes #23


release.js
```diff
@@ -36,15 +36,17 @@ module.exports = async function ({ github, context, inputs }) {
 
   const { opticUrl, npmTag, version, id } = releaseMeta
 
+  const run = runSpawn()
   if (!pr.merged) {
-    return github.rest.repos.deleteRelease({
+    await run('git', ['push', 'origin', '--delete', `"release/${version}"`])
+
+    return await github.rest.repos.deleteRelease({
```
simoneb, line 43: let's do these operations in a way that the failure of one doesn't cause the failure of the other. it would also be good if we could log, when something goes wrong, what went wrong

sameer-coder (approved):
lgtm

release.js
```diff
@@ -36,15 +36,31 @@ module.exports = async function ({ github, context, inputs }) {
...
-    return github.rest.repos.deleteRelease({
-      owner,
-      repo,
-      release_id: id,
-    })
+    const branchName = `release/${version}`
+    try {
+      await run('git', ['push', 'origin', '--delete', branchName])
+    } catch (err) {
+      core.setFailed(
```
simoneb, line 45: I don't know (and it's not clear from the docs) if calling setFailed also exits the process, let's return just in case it doesn't

test/release.test.js
```diff
@@ -89,11 +89,19 @@ tap.afterEach(() => {
 })
 
 tap.test('Should delete the release if the pr is not merged', async t => {
-  const { release } = setup()
+  const { release, stubs } = setup()
   const data = clone(DEFAULT_ACTION_DATA)
   data.context.payload.pull_request.merged = false
   await release(data)
 
+  t.ok(
```
simoneb, line 97: let's add tests for failure cases as well please

release.js
```diff
@@ -36,15 +36,31 @@ module.exports = async function ({ github, context, inputs }) {
...
+      await github.rest.repos.deleteRelease({
+        owner,
+        repo,
+        release_id: id,
+      })
+    } catch (err) {
+      core.setFailed(
+        `The release ${id} could not be deleted. Error: ${err.message}`
+      )
+    }
```
simoneb, line 60: I'm not sure this logic does what we think it does

release.js
```diff
@@ -36,15 +36,31 @@ module.exports = async function ({ github, context, inputs }) {
...
+    return
```
simoneb, line 61: this return is doing nothing

test/release.test.js
```diff
@@ -89,11 +89,19 @@ tap.afterEach(() => {
...
+    stubs.runSpawnStub.calledWith('git', [
```
simoneb, line 98: I realize you may have found this pattern in the code already, but these assertions are better done with https://sinonjs.org/releases/latest/assertions/. if you do them with tap's t.ok, when the assertion fails you won't get any details about why it failed but just a generic error message

ja-bravo: That's good to know, it would have made writing the tests much easier! I've updated the rest of the tests.

release.js
```diff
@@ -36,15 +36,30 @@ module.exports = async function ({ github, context, inputs }) {
...
-    return github.rest.repos.deleteRelease({
-      owner,
-      repo,
-      release_id: id,
-    })
+    const branchName = `release/${version}`
+    try {
+      await run('git', ['push', 'origin', '--delete', branchName])
+    } catch (err) {
+      return core.setFailed(
```
Eomm, line 45: Could we try to remove the release even if the branch delete fail?

release.js
```diff
@@ -36,15 +36,30 @@ module.exports = async function ({ github, context, inputs }) {
...
+      await github.rest.repos.deleteRelease({
+        owner,
+        repo,
+        release_id: id,
+      })
+    } catch (err) {
+      return core.setFailed(
+        `The release ${id} could not be deleted. Error: ${err.message}`
+      )
+    }
```
Eomm, line 60: If all works, we should return

```suggestion
    }
    return
```

simoneb: I had asked to remove the return but you're right, we can't continue. Why are the tests succeeding with this bug? We need more tests

release.js
ja-bravo: That was the original plan, I don't think it's a bad idea. What do you think @simoneb ?

release.js
ja-bravo: The tests were not checking that the function returned early. I have added 3 more tests to confirm this.

release.js
simoneb: yes

release.js
```diff
@@ -36,15 +36,36 @@ module.exports = async function ({ github, context, inputs }) {
...
-      release_id: id,
-    })
+    const branchName = `release/${version}`
+    try {
+      await run('git', ['push', 'origin', '--delete', branchName])
+    } catch (err) {
+      core.info(
+        `The branch ${branchName} could not be deleted. Will try to delete the release anyways.`
+      )
+      core.setFailed(
```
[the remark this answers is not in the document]
ja-bravo: As far as I can see, this only sets the exit code to failure and lets the action continue.
> You should use this library to set the failing exit code for your action. If status is not set and the script runs to completion, that will lead to a success.

I tested adding a ```core.setFailed``` right at the top of the ```release.js``` file and everything worked, the only change being that the action will be marked as failed. @simoneb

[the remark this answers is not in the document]
simoneb: okay good, but then what if both fail? shall we not use both failures to form the final error message? What about using Promise.allSettled instead?

Eomm (approved):

release.js
```diff
@@ -36,15 +36,37 @@ module.exports = async function ({ github, context, inputs }) {
...
+      logError(branchPromise.reason.message)
+      logError(releasePromise.reason.message)
+      core.setFailed(`Could not delete the branch or the release.`)
+    } else if (branchPromise.status === 'rejected') {
+      logError(branchPromise.reason.message)
+      core.setFailed(`Could not delete branch ${branchName}`)
+    } else if (releasePromise.status === 'rejected') {
+      logError(releasePromise.reason.message)
+      core.setFailed(`Could not delete release ${id}`)
+    }
```
simoneb, line 64: let's simplify this logic please:

- extract error messages, if any
- put them in an array
- if the array has elements, set the action failed with the joint error messages

ja-bravo: I have updated the logic and while it is simpler, I feel it gives the user a bit less info. Let me know what you think.

simoneb (approved):
LGTM, good job!

release.js
```diff
@@ -36,15 +36,31 @@ module.exports = async function ({ github, context, inputs }) {
...
+    const errors = promises.filter(p => p.reason).map(p => p.reason.message)
+    if (errors.length) {
+      core.setFailed(
+        `Something went wrong while deleting the branch or release. \n Errors: ${errors.join(
+          '\n'
+        )}`
+      )
+    }
+
+    // Both the branch and release have been deleted, return early
```
simoneb, line 60: minor: this comment is not accurate anymore now. something may have failed, but we return nonetheless

PR #30, opened 2021-12-09
feat: build phase

### LINKED ISSUE #27: Customize node version and custom build steps
The action is running npm install to install pacakges in https://github.com/nearform/optic-release-automation-action/blob/main/action.yml#L51.

This raises a few questions, e.g.:

- what Node version does it run on?
- what if the project uses a different package manager?
- what if the project using this action needs to do more things in order to generate a build artifact?

Overall it sounds like more control over what's run by this action is necessary

Eomm:
Fix #27 

This PR:

- refactor the action.yml running more little steps:
  - runs the project clone. The setup overwrite any `actions/checkout` run previously
  - runs the package version bump using a shell instead of a node.js script
  - runs an optional "build step"
  - runs the git commands and PR logic using a node.js script 
  - add the `git add` command to let the user build new artefacts during the build phase. _The previous `git commit -a` flag does not add new files to the staging, it allows adding only changed files._
  - release the package as it was done previously. Only this step needs the NPM_TOKEN.
  - removed the `npm install` phase: we don't know how to install the user's package. The installation command will be provided by the user by using the `build-command` option.
- tried to improve the doc clarity
- add the docs about how to build a module and release it
- add a better output log using `npm pack`: this will help the user to check what has been release

**Notes**

I think this PR is semver minor.


README.md
```diff
@@ -48,16 +55,106 @@ jobs:
...
+
+To do it you need to add some extra steps to your workflow.
+
+##### To be committed
+
+In this example, the `npm run build` command will create some artifacts that will be committed to the repository.
+Of course, you can use any build tool you need.
+_These artifacts should not be listed in the `.gitignore` file._
+
+The action will adds the changes in the new release PR for you!
```
ja-bravo, line 89:
```suggestion
The action will add the changes in the new release PR for you!
```

README.md
```diff
@@ -48,16 +55,106 @@ jobs:
...
+
+Now your PR includes the build artifacts, so when you will merge the PR, the action will publish the artifacts to Npm as expected.
+
+_Another approach is to commit manually the artifacts you want to release into the release branch._
+
+
+##### To be released
+
+You may want to release the artifacts without committing them to the repository such as a `dist/` folder.
+In this case, your workflow will look similar to the previous one, but the build step needs to run when the PS is merged:
```
ja-bravo, line 137:
```suggestion
In this case, your workflow will look similar to the previous one, but the build step needs to run when the PR is merged:
```

I think this is a typo, ignore it if I'm wrong

README.md
```diff
@@ -6,17 +6,24 @@ This action allows you to automate the release process of your npm modules. It c
...
 
-- Install the [optic-release-automation](https://github.com/apps/optic-release-automation) GitHub app to your organization (or selected repositories)
+- _(Optional)_ Install the [optic-release-automation](https://github.com/apps/optic-release-automation) GitHub app to your organization (or selected repositories)
 - Create a new workflow file at `.github/workflows/release.yml` (from example below) with one step that uses this action and supply the inputs.
 
-#### Example
+Note that the `on` triggers are mandatory:
+
+- `workflow_dispatch`: to start the new release process
+- `pull_request`: to complete the release process when the PR is merged
```
simoneb, line 22: since you're specifying this, you may want to specify that the only state we need is `closed`.

README.md
```diff
@@ -48,16 +55,106 @@ jobs:
...
+##### To be released
+
+You may want to release the artifacts without committing them to the repository such as a `dist/` folder.
+In this case, your workflow will look similar to the previous one, but the build step needs to run when the PR is merged:
+
+```yml
+      # [...] same workflow as before ...
+
+      - name: build the project when the PR is merged
+        if: ${{ github.event_name == 'pull_request' }}
```
simoneb, line 143: are you sure this condition is right? I mean this action creates the PR on the workflow_dispatch event, so the changes (e.g. creating a dist folder) need to happen *before* you create the PR.

README.md
```diff
@@ -48,16 +55,106 @@ jobs:
...
+        run: |
+          npm install
+          npm run build
```
simoneb, line 146: ok so now I have a doubt about the approach. let's say that we want to embed the version number *after* the bump inside the application, we might do that by reading the package.json file *after* the bump during in the build script. What that means is that the build script would need to be invoked *after* the bump is done. Because the bump is done I'm now wondering if this approach of externalizing npm i / npm run build outside the action simply wouldn't work, and instead we have to use the approach used by cypress (as described in the issue)

simoneb (commented):
See inline comments. I am starting to suspect this approach won't work.

README.md
Eomm: Using the documented workflow, the PR creation event is ignored as we filter by `type: closed`.

In this case, we want to release the `dist/` folder without committing the content.
When the user merges the PR (event `pull_request`, type `closed`) this action runs the `npm publish` command.
So whatever the user puts into the CWD will be packed and pushed to npm.

Creating the `dist/` folder before creating the PR would mean that:
- the user creates the build
- the action bump the version and commit it into a `release/$version` branch
- the action open a PR
- the workflow ends

So, in this case, the build would be pruned when the workflow end

README.md
Eomm: You are right, this case can be covered - I need to think a bit about how we could manage it for the user.

I may check deeper the cypress use case if there are some downside

Eomm:
> See inline comments. I am starting to suspect this approach won't work.

I strongly believe that the approach we are documenting is good 😄
After a lot of testing here https://github.com/Eomm/fastify-overview/actions/workflows/play.yml and I think this proposal is the most flexible to every project out there.

Flexible means that "standard" packages need to copy&paste the configuration and the "edge case" packages need to add a step to prepare the build.
I'm thinking about compiled packages that do not need to set a nodejs_version, as [sharp does](https://github.com/lovell/sharp/blob/master/.github/workflows/ci-darwin-arm64v8.yml)

I will try to compare it with the cypress logic to understand better what we are missing

Eomm:
@simoneb Here are my findings:

- stay away from the cypress approach. The build script runs with `node12`: no matter what. So there are many issues about this on their issue tracker https://github.com/cypress-io/github-action/issues/467
- we could ease the user to accomplish a complex task as the `version` use case you have mentioned but the user needs to write a simple commit script.

I draw a quick image flow. The grey boxes represent the users' custom steps.

- The upper boxes are already documented by this PR.
- The lower left box would support the "build with `version` use case".
- The lower right one must be ignored

I think the solution could be:

- define a GitHub output variable to let the user commit easily to the `release/$version` branch
- document this use case


optic-action


I would improve and include this image into the docs as well, it seems clearer to understand to me.

Do you agree?

simoneb:
@Eomm would this change be the way we recommend people to do it perhaps? https://github.com/nearform/optic-release-automation-action/pull/32/files#diff-d2bc4bbf14eadc292d84e0ef5f7a93115c23557f533809fc80b896934291529dR4 (or something similar)

Eomm:
> @Eomm would this change be the way we recommend people to do it perhaps? https://github.com/nearform/optic-release-automation-action/pull/32/files#diff-d2bc4bbf14eadc292d84e0ef5f7a93115c23557f533809fc80b896934291529dR4 (or something similar)

I would suggest that kind of command.
(I need to be sure that the current git branch is the one the action sets)

action.yml
```diff
@@ -44,11 +44,12 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: actions/checkout@v2
+    - name: checkout if not already done
+      if: hashFiles('package.json') == ''
+      uses: actions/checkout@v2.4.0
```
simoneb, line 49: I don't think we should be implementing this kind of implicit behavior. If a checkout is needed, we need to decide if we do it inside the action or if we need the user to do it before invoking the action. Having conditional behavior will lead to confusion and bugs most likely.

bump.js
```diff
@@ -47,7 +47,8 @@ module.exports = async function ({ context, inputs }) {
   const branchName = `release/${newVersion}`
 
   await run('git', ['checkout', '-b', branchName])
-  await run('git', ['commit', '-am', newVersion])
+  await run('git', ['add', '-A'])
+  await run('git', ['commit', '-m', newVersion])
```
simoneb, line 51: oh by the way, we need to make this customizable. if we don't have an issue, please open one. anything where we're making decisions we need to allow the user to customize the decision. in this case, we may use a formatting utility so that the user can provide a commit message containing a placeholder that we replace with the version

test/bump.test.js
```diff
@@ -78,7 +78,8 @@ tap.test('should create a new git branch', async t => {
   const branchName = `release/${TEST_VERSION}`
 
   t.ok(stubs.runSpawnStub.calledWith('git', ['checkout', '-b', branchName]))
-  t.ok(stubs.runSpawnStub.calledWith('git', ['commit', '-am', TEST_VERSION]))
+  t.ok(stubs.runSpawnStub.calledWith('git', ['add', '-A']))
```
simoneb, line 81: all these assertions should be replaced with sinon.assert. these assertions will only give a generic, non-meaningful error message when they fail

simoneb:
> Do you agree?

@Eomm I'm afraid this is really not clear to me. I may have a simplified view of the problem but really we need to make it easy for the user to understand AND implement custom build steps.

There is a single representative scenario I can think of, with these being the acceptance criteria:

- When I run the workflow containing this action:
  - I want to run custom "code" (e.g. npm run build) *after* the version bump is done and *before* the PR is created
  - I want to have the new version number available (meaning, already stored in package.json) when I run my custom code
  - Whatever I do in my custom code, if it generates files which are not .gitignored, I want those files to be committed and included as changes in the PR opened by this action
 - I don't need any custom code to be executed after the PR is merged, the behavior of the action is already good (it simply published the draft release)

We should try to avoid requiring the user to configure the action in fancy ways to obtain this. E.g. I don't expect the user to need to provide any custom git commit commands. As a user I would only like to provide the custom scripts that I need to be run

Eomm:
The PR is now ready for review 🎉

Here is a preview of this new implementation in action:
https://github.com/Eomm/node-workshop-micro/pull/5

I think this is a semver minor, even if the codebase changed a lot.

To support the build, the only thing the user needs to set up is the github runner by setting/installing whatever the application needs.

This implementation solves all the issues the cypress implementation has running the build command from a node.js script and letting the user creates complex build such as sharp:

- https://github.com/cypress-io/github-action/issues/467
- https://github.com/lovell/sharp/blob/master/.github/workflows/ci-darwin-arm64v8.yml

simoneb:
@Eomm please update the PR description which is now, I suspect, outdated

Eomm:
> @Eomm please update the PR description which is now, I suspect, outdated

Updated 👍🏽

simoneb:
If this is the way you'd do it, this is exactly what I had in mind https://github.com/Eomm/node-workshop-micro/blob/master/.github/workflows/optic-release.yml#L23

simoneb:
Why would this be a minor? If we stop doing npm install this will break existing usages

simoneb:
> Are you thinking about `pre/post` scripts? 

I'm thinking about anything that npm install that we were always doing before might have done. 

>In this case yes, we are skipping them now.

Why are we skipping them now?

package.json
```diff
@@ -2,10 +2,11 @@
   "name": "optic-release-automation-action",
   "version": "2.3.0",
   "description": "Automate the release process of your npm modules by providing them OTP on-demand.",
-  "main": "index.js",
+  "main": "action.js",
```
simoneb, line 5: we really need to restructure this repo moving all the source code to a src folder as we did in the merge action. do you mind taking the chance to do so in this PR please?

simoneb (commented):
Overall LGTM but a few comments, questions and changes needed.

action.yml
```diff
@@ -48,15 +51,47 @@ inputs:
...
+    - name: Commit and open PR
+      uses: actions/github-script@v5
+      if: ${{ github.event_name == 'workflow_dispatch' }}
+      with:
+        github-token: ${{ inputs.github-token }}
+        script: |
+          const script = require('${{ github.action_path }}/dist/index.js')
+          await script({ github, context, inputs: ${{ toJSON(inputs) }} })
+      env:
+        NPM_VERSION: ${{ steps.version-bump.outputs.NPM_VERSION }}
```
simoneb, line 88: ok no let's provide this as an input to the script please, not an env var which is inconsistent with the way we're providing other input

action.yml
```diff
@@ -48,15 +51,47 @@ inputs:
...
-        npm install
       shell: 'bash'
-    - uses: actions/github-script@v5
+
+    - name: Version bump
+      id: version-bump
+      if: ${{ github.event_name == 'workflow_dispatch' }}
+      run: |
+        npm version --no-git-tag-version ${{ inputs.semver }}
+        echo "::set-output name=NPM_VERSION::$(echo $(node -p "require('./package.json').version"))"
```
simoneb, line 68: let's not call this output variable NPM_VERSION, it's confusing. PACKAGE_VERSION perhaps?

simoneb:
It's also not clear to me why the husky scripts that are in place in some repos wouldn't already achieve this without changing the action. Is it because husky scripts are not run in CI?

README.md
```diff
@@ -53,36 +60,44 @@ jobs:
...
 
-The above workflow (when manually triggered) will
+The above workflow (when manually triggered ⚡️) will:
 
-- Run `npm version <semver>` command to bump the version as configured (patch, minor, etc)
-- Open a PR that looks like following
+- 🤖 Run `npm version <semver>` command to bump the version as configured (patch, minor, etc)
+- 🤖 Execute the `build-command` if configured
+- 🤖 Commit the changes and push to the `release/${new semver version}` branch
+- 🤖 Open a PR that looks like following
```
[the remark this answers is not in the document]
Eomm: I aimed to give the user a fast overview of the actions done by this GHA.

Do you find it odd?

package.json
Eomm: I thought the same, but I was afraid to add confusion for the reviewers

Eomm:
> Why are we skipping them now?

Calling `npm install` is wrong. Removing this command will skip all the[ `install`'s related events.](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install)

> It's also not clear to me why the husky scripts that are in place in some repos wouldn't already achieve this without changing the action. Is it because husky scripts are not run in CI?

And what about those repositories that do not have the husky script?

On a local env, I think the husky hook is weak because you can skip it with the `git commit -n` flag.

If a user or the CI install the module using the `npm I --no-script` flag the husky hook won't be installed too

package.json
simoneb: can you create an issue for this? we'll do it on the next round, I agree it would be hard to spot changes if we did it in this PR

openPr.js
```diff
@@ -41,18 +41,19 @@ const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
...
 
-  const newVersion = await run('npm', [
-    'version',
-    '--no-git-tag-version',
-    inputs.semver,
-  ])
+  if (!process.env.NPM_VERSION) {
+    throw new Error('Env var NPM_VERSION is missing!')
+  }
+  const newVersion = `v${process.env.NPM_VERSION}`
```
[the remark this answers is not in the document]
simoneb: ok I see now, but I would still prefer to get is as a function argument for consistency, see my other comment about this

README.md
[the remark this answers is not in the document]
simoneb: yes. apologies but I don't personally like playful things in docs

simoneb:
> > Why are we skipping them now?
> 
> Calling `npm install` is wrong. Removing this command will skip all the[ `install`'s related events.](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install)
> 

Okay gotcha. When you say that npm install is wrong you mean that it's wrong to do it within the action? If so, I tend to agree, e.g. because you're using a different package manager. So fine, let's assume that the user is providing npm install in the new action input you added in this PR

> > It's also not clear to me why the husky scripts that are in place in some repos wouldn't already achieve this without changing the action. Is it because husky scripts are not run in CI?
> 
> And what about those repositories that do not have the husky script?
> 
> On a local env, I think the husky hook is weak because you can skip it with the `git commit -n` flag.
> 
> If a user or the CI install the module using the `npm I --no-script` flag the husky hook won't be installed too

Okay but you're not seeing my point of view here. As a user, I want to understand what options I have to create that dist folder. If I have husky and if husky pre-commit runs npm run build, can I avoid to explicitly invoke it in the script I provide as `build-command` input? If so, as a user I should know so that I avoid building it twice!

Eomm:
I have addressed all the new feedback

> So fine, let's assume that the user is providing npm install in the new action input you added in this PR

Now I get why you suggest this is a breaking change: because of those repositories that were using husky and they were relying on the `npm install` execution by this GHA.

In this case, I think there is not much we can do because:

- the `npm install` was a bug, not a feature: we can assume that nobody is using this GHA with `yarn` actually
- the docs were suggesting to use the `@main` branch, so we can't even bump a major. We should work on a multi branches repository to support this use case

https://github.com/nearform/optic-release-automation-action/blob/main/.github/workflows/release.yml#L17

simoneb (approved):
Very hard to see if anything could break here. Overall looks good. Let's try it out and see what happens

Issue #26, opened 2021-12-07
PR creation fails when body too long


simoneb:
The PR creation fails when the body is too long due to an error in the github rest api, this may happen for example when trying to do a first release with many changes in it.

The limits don't seem to be clearly documented in the api docs but we should handle that situation as it leads to an inconsistent state

PR #31, opened 2021-12-09
fix: error handling in callApi method

TomasTM:
closes: #26


utils/callApi.js
```diff
@@ -1,27 +1,28 @@
...
+      },
+      body: JSON.stringify(body),
+    })
 
-  if (response.status !== 200) {
+    return response.json()
+  } catch (error) {
     logWarning(`Please ensure that Github App is installed ${GITHUB_APP_URL}`)
+    logError(error)
   }
```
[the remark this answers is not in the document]
simoneb: @ja-bravo is right. fetch does not throw on http response errors.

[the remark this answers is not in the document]
TomasTM: Sheesh, you're right! not sure we need to throw an error I guess log error would be enough right?

simoneb (changes requested):
This needs a bit more work and thought put into it. As you can see currently the code assumes that if a non-200 error comes back, it's because the GitHub app which works in the background of this action is not installed.

That's surely one of the most common reasons why errors would happen, but not the only reason! So we need to make this error detection logic more robust.

In the case we're trying to handle here, the GH API was returning a legitimate error because the body of the PR we were trying to create was too large.

As a user, hence, I would expect:

- the workflow containing the action to fail (see actions/toolkit for helpers to do that)
- a generic error message suggesting that a probable cause might be that the app is not installed (similar to what we have now)
- an error message to be logged showing more details. note that logging the response object is not a good idea, we should log only the body of the response or in general whatever may be relevant for the user
- any operations that the action may have done to be undone (e.g. if it created and pushed a branch, it should undo it)

All of this should come with tests

TomasTM:
@simoneb I think I addressed all the requests you've made

utils/callApi.js
```diff
@@ -19,6 +19,7 @@ const callApi = async ({ method, endpoint, body }, inputs) => {
 
   if (response.status !== 200) {
     logWarning(`Please ensure that Github App is installed ${GITHUB_APP_URL}`)
+    return response
```
simoneb, line 22: why are you returning the response? are we using it?

bump.js
```diff
@@ -62,18 +63,22 @@ module.exports = async function ({ context, inputs }) {
...
+          body: prBody,
+        },
       },
-    },
-    inputs
-  )
+      inputs
+    )
+  } catch (err) {
+    await run('git', ['push', 'origin', '--delete', branchName])
```
simoneb, line 81: now this is more of a detail and surely we're not being so specific about error handling in other places, but for the sake of completeness it should be noted that this operation may also fail, in which case we wouldn't give any details to the user.

utils/callApi.js
simoneb: why would you return it?

TomasTM: to make sure no actions continue if not 200

simoneb: sounds good

simoneb:
please link the issue properly

utils/callApi.js
ja-bravo, line 22: Wouldn't this tell the user that every error is about the app not being installed? Not sure if the server returns useful info about errors but if it does, I'd be good to log them, I think.

TomasTM: I guess the `setFailed` will give the user feedback per action failure, right? we can logError as well but I think it may be a bit redundant

simoneb: You're right @ja-bravo that this warning we're logging is really generic, that's certainly something to improve for the future but not really in scope for this PR.

bump.js
```diff
@@ -62,18 +63,26 @@ module.exports = async function ({ context, inputs }) {
...
-  )
+      inputs
+    )
+  } catch (err) {
+    try {
+      await run('git', ['push', 'origin', '--delete', branchName])
+    } catch (error) {
+      core.setFailed(`Unable to create the pull request ${error.message}`)
+    }
+    core.setFailed(`Unable to create the pull request ${err.message}`)
```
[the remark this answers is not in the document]
TomasTM: @simoneb I did have a typo in the first `setFailed` it should be `core.setFailed(`Unable to delete the pull request ${error.message}`)` but this should do what you described. if only one of those actions fails only one message will be shown to the user, if both actions fail both errors will be printed.

bump.js
```diff
@@ -68,18 +69,27 @@ module.exports = async function ({ context, inputs }) {
...
-    inputs
-  )
+      inputs
+    )
+  } catch (err) {
+    let message = `Unable to create the pull request ${err.message}`
+    try {
+      await run('git', ['push', 'origin', '--delete', branchName])
+    } catch (error) {
+      message += `\n Unable to delete the pull request ${error.message}`
```
simoneb, line 91: this message is wrong

TomasTM: Opss, fixed

simoneb (approved):

PR #34, opened 2021-12-13
Feat: Customize commit message

### LINKED ISSUE #33: Customize commit message
The user should be able to configure the commit message performed by this action when the version is bumped.

oh by the way, we need to make this customizable. if we don't have an issue, please open one. anything where we're making decisions we need to allow the user to customize the decision. in this case, we may use a formatting utility so that the user can provide a commit message containing a placeholder that we replace with the version

_Originally posted by @simoneb in https://github.com/nearform/optic-release-automation-action/pull/30#r767186129_

ja-bravo:
This PR allows the user of the action to customize the commit message using a template.

### Usage
When running the app, the user can change the default message template using the workflow_dispatch inputs. The keyword {version} will be replaced by the new version being commited.

Closes #33


README.md
```diff
@@ -94,6 +99,7 @@ jobs:
 | ---            | ---      | ---                                                                                                                                                                                        |
 | `github-token` | Yes      | This is your GitHub token, it's [already available](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) to your GitHub action |
 | `semver`       | Yes      | The version you want to bump (`patch|minor|major`).                                                                                                                                        |
+| `commit-message`       | Yes      |The commit message template. The keyword "{version}" will be replaced with the new version.                                                                                                                                        |
```
Eomm, line 102:
```suggestion
| `commit-message`       | No      |The commit message template. The keyword `{version}` will be replaced with the new version.                                                                                                                                        |
```

ja-bravo: Nice catch

bump.js
```diff
@@ -46,8 +47,13 @@ module.exports = async function ({ context, inputs }) {
   ])
   const branchName = `release/${newVersion}`
 
+  const messageTemplate = inputs['commit-message'] || 'Release {version}'
```
simoneb, line 50: the default will come from the action definition, we don't have to be redundant here

simoneb (approved):
One comment, LGTM otherwise

bump.js
ja-bravo: That was there because the commit-message input is not required so the user could just leave a white space. Maybe that's okay and we can remove this check? 

Or we could set the commit-message input as required and the user can just use the default value if they don't want to modify it like the semver input.

https://github.com/nearform/optic-release-automation-action/blob/a60f257963910ea7c3d7a389accbe7abbb506a87/README.md#L35-L38

What do you think @simoneb ?

simoneb: well then we have an inconsistency between readme and the action definition https://github.com/nearform/optic-release-automation-action/pull/34/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R45

ja-bravo: I have updated the documentation and behaviour to make the input not required and fall back to the default value in the action definition

simoneb (approved):

package-lock.json
```diff
@@ -1,84 +1,57 @@
 {
   "name": "optic-release-automation-action",
   "version": "2.3.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
Eomm, line 4: Could you add the `.nvmrc` file?
We are running node12 here and the previous lock file was wrong. These changes are good!

Eomm (approved):

PR #37, opened 2021-12-14
fix: build step

TomasTM:


simoneb:
I don't know what changes are in this but we need to make sure we're committing the right dist folder when we push commits, which should happen automatically thanks to the Husky scripts. @TomasTM do you know why that didn't happen in your latest PR?

TomasTM:
@simoneb sorry this was indeed updated correctly, @Eomm just confirmed 😄  - I'll delete this pr

PR #45, opened 2021-12-21
Add release author to PR

### LINKED ISSUE #40: Add author information
Looking at a release PR: https://github.com/nearform/optic-release-automation-action/pull/39

you can't know who has started the release.
It could be useful for maintainers knowing it to sync offline

sameer-coder:
This PR mentions the user who triggered the automation workflow as the release author in the PR template.

Closes #40

### DIFF
--- a/openPr.js
+++ b/openPr.js
@@ -35,6 +35,7 @@ const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
     tagsToUpdate: tagsToBeUpdated.join(', '),
     npmPublish: !!inputs['npm-token'],
     syncTags: /true/i.test(inputs['sync-semver-tags']),
+    author: inputs['actor-name'],
   })
 }
 
--- a/openPr.js
+++ b/openPr.js
@@ -14,7 +14,7 @@ const transformCommitMessage = require('./utils/commitMessage')
 const actionPath = process.env.GITHUB_ACTION_PATH
 const tpl = fs.readFileSync(path.join(actionPath, 'pr.tpl'), 'utf8')
 
-const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
+const getPRBody = (template, { newVersion, draftRelease, inputs, author }) => {
   const tagsToBeUpdated = []
   const { major, minor } = semver.parse(newVersion)
 
@@ -35,7 +35,7 @@ const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
     tagsToUpdate: tagsToBeUpdated.join(', '),
     npmPublish: !!inputs['npm-token'],
     syncTags: /true/i.test(inputs['sync-semver-tags']),
-    author: inputs['actor-name'],
+    author,
   })
 }
 
@@ -70,7 +70,12 @@ module.exports = async function ({ context, inputs, packageVersion }) {
     inputs
   )
 
-  const prBody = getPRBody(_template(tpl), { newVersion, draftRelease, inputs })
+  const prBody = getPRBody(_template(tpl), {
+    newVersion,
+    draftRelease,
+    inputs,
+    author: context.actor,
+  })
   try {
     await callApi(
       {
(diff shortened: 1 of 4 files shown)


openPr.js
```diff
@@ -35,6 +35,7 @@ const getPRBody = (template, { newVersion, draftRelease, inputs }) => {
     tagsToUpdate: tagsToBeUpdated.join(', '),
     npmPublish: !!inputs['npm-token'],
     syncTags: /true/i.test(inputs['sync-semver-tags']),
+    author: inputs['actor-name'],
```
Eomm, line 38: Could we read the author from the GH context instead?
This requires to the user to set this info if I'm not wrong

sameer-coder: The user doesn't need to set this info as we use this default in action.yml

```yaml
  actor-name:
    description: "The name you want to see in the new release commit."
    required: false
    default: "${{ github.actor }}"
```

But yes, I too initially though of using the value from the `context` as that would be cleaner approach.

salmanm: I guess we initially added this `actor-name` input without much thought. But I think it wouldn't be so useful to let user modify the actor name so perhaps we could get rid of this input and use `github.actor` everywhere? /cc @simoneb

sameer-coder: I have updated the PR to use the `actor-name` from `context`

Eomm (approved):

pr.tpl
```diff
@@ -4,6 +4,8 @@ This **draft** PR is opened by Github action [optic-release-automation-action](h
 
 A new **draft** GitHub release [${releaseMeta.version}](${draftRelease.html_url}) has been created.
 
+Release author: [${author}](https://github.com/${author})
```
simoneb, line 7: you're not mentioning the author in this way

```suggestion
Release author: @${author}
```

simoneb:
Okay but what about the actor-name input? I mean if we have that input, wouldn't we expect this new value to change in the PR body as well if we provide it?

sameer-coder:
> Okay but what about the actor-name input? I mean if we have that input, wouldn't we expect this new value to change in the PR body as well if we provide it?

That's really the question, if we should allow the user to manually change `actor-name` and `actor-email`. If we decide to keep these options then I will revert to my earlier implementation which used the `actor-name` from the `inputs`. After reading Salman's comment I too thought that it might not be very helpful to let user modify the actor name and that we could get rid of these options.

simoneb:
Okay but in this way we have a mixed solution. Is this what we want?

sameer-coder:
No, I can remove `actor-name` and `actor-email` from the inputs so that we only use them from the github context. I was waiting for your review. I will make the change now.

simoneb:
Are you sure we have the email in the context? (I don't know)

sameer-coder:
No, we don't have email in context. Should we use the default `actions@users.noreply.github.com` or something like `<actor-name>@users.noreply.github.com` 🤔

simoneb (commented):
I think that within the context of this PR we should also take the chance to figure out if  we can make it so that whoever triggers the release workflow is also the person on behalf of which we commit the changes.

I am not actually 100% sure if this can work all the time, but since we have the username in the context, why don't we make an API call to retrieve that user's details and use those when configuring git before we commit?

It may require changes to the flow of the action but I think it's something worth investigating. Note that the best way to try this out is to use a dummy repo which references the branch of this repo where we have these changes

Eomm:
The `actor-name` and `actor-email` params are used to set the commit information, so a user may set some email group.

> I think that within the context of this PR we should also take the chance to figure out if we can make it so that whoever triggers the release workflow is also the person on behalf of which we commit the changes.

Maybe it is doable by suggesting to the user how to set the `actor-name` and `actor-email` params (as it has been done for the multi-user OTP).

---

The scope of the initial issue I opened was to know who pushed the button to start a release, for this reason, I suggested reading the `context`'s author.

simoneb (approved):
LGTM. Let's keep this change limited to what the original issue was requesting

PR #46, opened 2021-12-21
docs: use choice type for the semver input

### LINKED ISSUE #42: Use choice workflow input type to get the semver bump
As seen in https://github.com/MylesBorins/node-osc/blob/main/.github/workflows/bump-version.yml#L7.

It may slightly improve the user experience

ja-bravo:
Update the docs to use the choice type for the semver input. No changes are needed to the actual action.

Wondering if it'll be a good idea to add another choice input to publish to NPM. Mantainers may want to not publish one release to NPM without having to update the `release.yml` or remove the secret.

Closes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,11 @@ on:
         description: "The semver to use"
         required: true
         default: "patch"
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
       tag:
         description: "The npm tag"
         required: false


salmanm (approved):
Nice! I didn't notice GH added the `choice` type recently.

simoneb (commented):
Can you please also change the release workflow in this repo to use it, so we can actually confirm that it works?

simoneb (approved):

PR #52, opened 2022-01-05
chore: add simple diff tool in gitconfig file

### LINKED ISSUE #41: Add proper .gitconfig
We should add a proper `.gitconfig` to this repository to avoid this effect:

https://github.com/nearform/optic-release-automation-action/pull/39/files#diff-3d2b59189eeedc2d428ddd632e97658fe310f587f7cb63b01f9b98ffc11c0197

The line ending has changed

TomasTM:
closes #41 

Implement basic .gitconfig file to better manage line endings

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,4 @@
+[diff]
+	tool = vimdiff
+[difftool]
+	prompt = false
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,6 +1,4 @@
 [core]
 	autocrlf = input
-[diff]
-	tool = vimdiff
 [difftool]
 	prompt = false
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,3 +1,5 @@
+[core]
+	autocrlf = input
 [diff]
 	tool = vimdiff
 [difftool]


TomasTM:
@Eomm we can get pretty fancy with the config but this should be the bare minimum. let me know if you want to go crazy with [delta] or [color]

.gitconfig
```diff
@@ -0,0 +1,4 @@
+[diff]
```
Eomm, line 1:
```suggestion
[core]
	autocrlf = input
[diff]
```

Eomm (commented):
for that issue, I was thinking more about the line endings.

the build executed by this action has committed an unchanged file:

https://github.com/nearform/optic-release-automation-action/pull/39/files#diff-3d2b59189eeedc2d428ddd632e97658fe310f587f7cb63b01f9b98ffc11c0197

modifying only the line endings, so I suppose it is an `autocrlf` issue.

The other `diff*` settings are more personal in my opinion

.gitconfig
```diff
@@ -0,0 +1,6 @@
+[core]
+	autocrlf = input
+[diff]
+	tool = vimdiff
+[difftool]
+	prompt = false
```
simoneb, line 6: are these needed? I have concerns they're portable cross platform

TomasTM: Humm yeah that is a good point! I'll remove the tool and leave the prompt so that the confirmation doesn't appear

simoneb:
Can we please make sure to **always** describe what the PR does and link the issue please?

Eomm (approved):

simoneb (changes requested):
blocking until we have a PR description and a linked issue

simoneb (approved):

PR #54, opened 2022-01-07
fix: tags syncing

salmanm:
Fixes https://github.com/nearform/optic-release-automation-action/issues/53

### DIFF
--- a/.github/workflows/check-linked-issues.yml
+++ b/.github/workflows/check-linked-issues.yml
@@ -1,0 +1,16 @@
+name: Check Linked Issues
+
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize]
+
+jobs:
+  check_pull_requests:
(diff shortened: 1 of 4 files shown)


simoneb (commented):
I'm not sure I understand the logic here. Do we actually need all these exceptions?

I mean, why is 0 a special case? Major 0 is a valid version, you can release version v0.0.1, and in that case I would expect to have both all of:

- v0
- v0.0
- v0.0.1

And I think this applies to all version numbers in the same way. So my question is, can't this logic actually get much simpler than it is now?

salmanm:
You're right, but I hadn't removed the condition as I thought it's put there for a reason. But simplified now.

simoneb:
Yeah I think it was a lack of clear specifications since the beginning.

simoneb (approved):

PR #56, opened 2022-01-11
fix: ignore branch deleting errors

### LINKED ISSUE #55: On PR close, don't error when deleting branch if branch does not exist
When a release PR is closed manually, we try to delete the corresponding branch, which is correct.

Nonetheless the user who closed the PR may have already deleted the branch, because there's a convenient "Delete branch" button in the GitHub UI.

Hence, we should be able to support this scenario by avoiding to fail the action run if the branch is no longer there.

jhonrocha:
Ignoring errors when deleting the release branch.

Fixes #55 .

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10847,12 +10847,10 @@ module.exports = async function ({ github, context, inputs }) {
       }),
     ])
 
-    const errors = promises.filter(p => p.reason).map(p => p.reason.message)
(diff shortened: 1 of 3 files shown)


salmanm (approved):

simoneb (commented):
I think this could have been more fine grained and only ignore errors in case the branch didn't exist, instead of all errors

PR #69, opened 2022-02-28
Log enhancement

### LINKED ISSUE #68: Better logging
We need to improve the logging in general in this action.

Case Scenario: 
Sometimes we are dealing with projects that needs to be published in `npm` and when there is no `npm-token` the package will not be published but a new version will be released in GitHub we need to improve the log warning us about it.

ramonmulia:
Closes #68

### DIFF
--- a/src/release.js
+++ b/src/release.js
@@ -7,7 +7,7 @@ const core = require('@actions/core')
 const { callApi } = require('./utils/callApi')
 const { tagVersionInGit } = require('./utils/tagVersion')
 const { runSpawn } = require('./utils/runSpawn')
-const { logError, logInfo } = require('./log')
+const { logError, logInfo, logWarning } = require('./log')
 
 module.exports = async function ({ github, context, inputs }) {
   logInfo('** Starting Release **')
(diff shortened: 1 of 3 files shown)


src/release.js
```diff
@@ -75,6 +80,8 @@ module.exports = async function ({ github, context, inputs }) {
     } else {
       await run('npm', ['publish', '--tag', npmTag])
     }
+  } else {
+    logInfo('missing npm-token')
```
radomird, line 84: Why not use `logWarning` instead?

src/release.js
```diff
@@ -19,6 +20,7 @@ module.exports = async function ({ github, context, inputs }) {
     pr.user.login !== 'optic-release-automation[bot]' ||
     !pr.title.startsWith(PR_TITLE_PREFIX)
   ) {
+    logInfo('skipping release.')
```
radomird, line 23: Why not use `logWarning`  instead?

src/release.js
```diff
@@ -86,6 +93,8 @@ module.exports = async function ({ github, context, inputs }) {
       await tagVersionInGit(`v${major}`)
       await tagVersionInGit(`v${major}.${minor}`)
       await tagVersionInGit(`v${major}.${minor}.${patch}`)
+    } else {
+      logInfo('missing sync-semver')
```
radomird, line 97: I think we should use `logError` here if something as crucial as semver is missing.

PR #80, opened 2022-03-20
chore: fix log level and readability

Eomm:
If the user does not set the `sync-semver` option, it is shown as an error, but it is not an error tho.


simoneb (approved):

src/release.js
```diff
@@ -94,7 +94,7 @@ module.exports = async function ({ github, context, inputs }) {
       await tagVersionInGit(`v${major}.${minor}`)
       await tagVersionInGit(`v${major}.${minor}.${patch}`)
     } else {
-      logError('missing sync-semver')
+      logWarning('missing sync-semver')
```
simoneb, line 97: why are we logging this in the first place? I mean if an input is not set, which is not mandatory, we shouldn't care really

simoneb (approved):

PR #82, opened 2022-03-23
Delete release branch when npm package is successfully published

### LINKED ISSUE #79: Delete branch on success
Right now the `release/` branch deletion is performed only when the OPTIC's PR is not merged:

https://github.com/nearform/optic-release-automation-action/blob/a207fd4013fa12996d492cb9809e4e3411a2eddf/src/release.js#L42

Why this setting?
I think it should be good to delete it in every case to clean up the repository's branches.
The GH release is going to track the version using the git tag.

Example:

- branch not deleted after the release: https://github.com/fastify/fastify-accepts/tree/release/v2.2.0
- the tag created by the GH release: https://github.com/fastify/fastify-accepts/tree/v2.2.0

So, I think that keeping the branches nice and clean is good. What do you think?

castarco:
Fixes #79

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10843,24 +10843,29 @@ module.exports = async function ({ github, context, inputs }) {
   const { opticUrl, npmTag, version, id } = releaseMeta
 
   const run = runSpawn()
-  if (!pr.merged) {
-    const branchName = `release/${version}`
+  const branchName = `release/${version}`
 
+  try {
+    // We "always" delete the release branch, if anything fails, the whole
+    // workflow has to be restarted from scratch.
     logInfo(`deleting ${branchName}`)
+    await run('git', ['push', 'origin', '--delete', branchName])
+  } catch (err) {
+    // That's not a big problem, so we don't want to mark the action as failed.
+    logWarning('Unable to delete the release branch')
+  }
 
-    const [, deleteRelease] = await Promise.allSettled([
-      run('git', ['push', 'origin', '--delete', branchName]),
-      github.rest.repos.deleteRelease({
+  if (!pr.merged) {
+    try {
+      await github.rest.repos.deleteRelease({
         owner,
         repo,
         release_id: id,
-      }),
-    ])
-
-    // Verify any errors deleting the Release. Ignore minor issues deleting the branch
-    if (deleteRelease.reason) {
+      })
+    } catch (reason) {
+      // Verify any errors deleting the Release. Ignore minor issues deleting the branch
       core.setFailed(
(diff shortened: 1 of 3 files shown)


castarco:
ok, I messed up with a rebase, some changes I made to release.js ended up in index.js, fixing that...

dist/index.js
```diff
@@ -10843,9 +10843,8 @@ module.exports = async function ({ github, context, inputs }) {
   const { opticUrl, npmTag, version, id } = releaseMeta
 
   const run = runSpawn()
+  const branchName = `release/${version}`
   if (!pr.merged) {
-    const branchName = `release/${version}`
-
     logInfo(`deleting ${branchName}`)
```
simoneb, line 10848: do we need to keep the deletion here too?

castarco: We've been discussing with @ramonmulia , and most likely we'll delete branches always, independently of whether anything fails.

This approach was prior to the discussion (trying to minimise the need for changes on current tests and workflows). I'm making changes right now.

ramonmulia (approved):

src/release.js
```diff
@@ -39,24 +39,29 @@ module.exports = async function ({ github, context, inputs }) {
   const { opticUrl, npmTag, version, id } = releaseMeta
 
   const run = runSpawn()
-  if (!pr.merged) {
-    const branchName = `release/${version}`
+  const branchName = `release/${version}`
+
+  try {
+    // We only delete the branch in case the release was successful
```
simoneb, line 45: I'm not sure I understand this comment. at this stage nothing has happened yet, so unless I'm missing something we are not actually deleting the branch only if the release is successful, we are doing it always

src/release.js
```diff
@@ -39,24 +39,29 @@ module.exports = async function ({ github, context, inputs }) {
...
-    // Verify any errors deleting the Release. Ignore minor issues deleting the branch
-    if (deleteRelease.reason) {
+      })
+    } catch (reason) {
+      // Verify any errors deleting the Release. Ignore minor issues deleting the branch
       core.setFailed(
-        `Something went wrong while deleting the release. \n Errors: ${deleteRelease.reason.message}`
+        `Something went wrong while deleting the release. \n Errors: ${reason.message}`
       )
     }
```
simoneb, line 67: the commend at the next line is no longer accurate

src/release.js
castarco: Sorry, I rebased with a change that I forgot to push before.

simoneb (approved):

simoneb:
I'll merge this to see if this wake gh up

PR #85, opened 2022-03-23
feat: Revert commit when it fails to release

### LINKED ISSUE #78: feat: revert commit on failed release
When the merge is completed, if the release process is not successful, the release branch is not deleted.

I think this action should revert the merge commit if the release fails.
Then, the `release/*` branch should be deleted in any case.

Here is the case:

- successful merge https://github.com/fastify/fastify-accepts/pull/120
- failed to release https://github.com/fastify/fastify-accepts/actions/runs/2012543070 (due to wrong action setup)
- then try to re-run the release process and the failure due to the branch name https://github.com/fastify/fastify-accepts/actions/runs/2012564358

ramonmulia:
Closes #78


simoneb:
not sure what's up with this PR but CI is stuck

simoneb:
@castarco can you review please?

castarco (approved):
LGTM 👍🏻 . We'll have to wait until GH is able to run the CI pipeline though (also stuck in other PRs).

src/release.js
```diff
@@ -66,22 +67,31 @@ module.exports = async function ({ github, context, inputs }) {
...
+        `//registry.npmjs.org/:_authToken=${inputs['npm-token']}`,
+      ])
+
+      await run('npm', ['pack', '--dry-run'])
+      if (opticToken) {
+        const otp = await run('curl', ['-s', `${opticUrl}${opticToken}`])
+        await run('npm', ['publish', '--otp', otp, '--tag', npmTag])
+      } else {
+        await run('npm', ['publish', '--tag', npmTag])
+      }
```
simoneb, line 84: since gh is stuck and it will require another commit to trigger CI, why don't we take the chance to extract this inline logic in functions? it will make it easier to make sense of the logical flow of the code

simoneb (approved):
LGTM, see inline comment for an opportunity to refactor

PR #87, opened 2022-03-23
Skip npm publish when already done in previous failed workflow

### LINKED ISSUE #86: The release workflow is not atomic nor idempotent
Right now, the release workflow is not atomic nor idempotent. It's possible that everything goes well for the most part, the npm package gets correctly published, and then the action ends up failing when it tries to create a "new release" in Github.

### Proposed solution:
In order to mitigate this problem, I propose to check if the npm package has been already published before even trying yo publish it. This way, if the previous described scenario happens in real life, we can re-run this action, and it will be able to skip the npm publication and directly jump into the GH release creation, allowing us to fix "unfinished" releases.

castarco:
### Main changes
It skips NPM package publication if the package was already published in a previously failed workflow execution. This can happen if the NPM publication is successful, but the Github release fails afterwards.

This change allows us to re-run this workflow again without having to bump versions, so we can fix the "inconsistency" between GH releases and NPM packages' versions.

### Minor changes
The last PR merge introduced some "regressions" in some tests: some assertions validated that some commands were never triggered, but after having extracted the `publishToNpm` function that was never going to happen in these tests, because these steps were abstracted away into another function that was being stubbed. I replaced these assertions by a "stronger" one, related to the newly introduced function.

### Notes
In the future, this might need refactoring: to reorder some steps (in case any of them are idempotent), to simplify the code (i.e.: we could avoid having to perform some checks, and remove some if-else blocks).

Fixes #86

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10880,7 +10880,7 @@ module.exports = async function ({ github, context, inputs }) {
     const npmToken = inputs['npm-token']
 
     if (npmToken) {
-      await publishToNpm({ npmToken, opticToken, opticUrl, npmTag })
+      await publishToNpm({ npmToken, opticToken, opticUrl, npmTag, version })
     } else {
       logWarning('missing npm-token')
     }
@@ -10991,9 +10991,16 @@ module.exports = transformCommitMessage
 
 "use strict";
 
+
 const { runSpawn } = __nccwpck_require__(2137)
 
-async function publishToNpm({ npmToken, opticToken, opticUrl, npmTag }) {
+async function publishToNpm({
+  npmToken,
+  opticToken,
+  opticUrl,
+  npmTag,
+  version,
+}) {
   const run = runSpawn()
 
   await run('npm', [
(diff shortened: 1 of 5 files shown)


simoneb:
what's the status of this PR @castarco ?

castarco:
Sorry, I missed that notification @simoneb .

I'm not really sure about it, I could fix the conflicts and merge it, as it would make this action a bit more resilient... But as you said in the related issue description, maybe it's worth checking the other steps to see if any of them is idempotent, so we can reorder the execution flow to have simpler code.

dist/index.js
```diff
@@ -10991,9 +10991,16 @@ module.exports = transformCommitMessage
 
 "use strict";
 
+
```
castarco, line 10994: This file is auto-generated 😛 , I did not introduce this space here.

src/release.js
```diff
@@ -76,7 +76,7 @@ module.exports = async function ({ github, context, inputs }) {
     const npmToken = inputs['npm-token']
 
     if (npmToken) {
-      await publishToNpm({ npmToken, opticToken, opticUrl, npmTag })
+      await publishToNpm({ npmToken, opticToken, opticUrl, npmTag, version })
```
castarco, line 79: This new "parameter" is introduced to check for previous package publications.

test/release.test.js
```diff
@@ -159,11 +167,7 @@ tap.test(
       stubs.coreStub.setFailed,
       `Something went wrong while deleting the release. \n Errors: Something went wrong in the release`
     )
-    sinon.assert.neverCalledWith(stubs.runSpawnStub, 'npm', [
-      'publish',
-      '--tag',
-      'latest',
-    ])
+    sinon.assert.notCalled(stubs.publishToNpmStub)
```
castarco, line 170: The previous assert was not "strong" enough, it would always succeed after the introduction of the `publishToNpm` function in a previous PR.

rp4rk (approved):
Looks good, nice test coverage

simoneb:
@castarco please do a new release of the action by triggering the release workflow from the actions tab. either patch or minor bump

PR #113, opened 2022-05-11
fix: add not published check

### LINKED ISSUE #110: First release of a package on npm does not work
See https://github.com/nearform/graphql-auto-federate/actions/runs/2282819711

ianlovell-nearform:
- Sets allow publish if not published (404) error returned from `npm view --json`

resolves #110

### DIFF
--- a/src/utils/publishToNpm.js
+++ b/src/utils/publishToNpm.js
@@ -2,6 +2,38 @@
 
 const { runSpawn } = require('./runSpawn')
 
+async function allowNpmPublish(version) {
+  const run = runSpawn()
+
+  // We need to check if the package was already published. This can happen if
+  // the action was already executed before, but it failed in its last step
+  // (GH release).
+  let packageName = null
+  try {
+    const packageInfo = await run('npm', ['view', '--json'])
+    packageName = packageInfo ? JSON.parse(packageInfo).name : null
+  } catch (error) {
+    if (!error?.message?.match(/code E404/)) {
+      throw error
+    }
+  }
+
+  // Package has not been published before
+  if (!packageName) {
+    return true
+  }
+
+  // NPM only looks into the remote registry when we pass an explicit
+  // package name & version, so we don't have to fear that it reads the
+  // info from the "local" package.json file.
+  const packageVersionInfo = await run('npm', [
+    'view',
+    `${packageName}@${version}`,
+  ])
+
+  return packageVersionInfo === ''
+}
+
 async function publishToNpm({
   npmToken,
   opticToken,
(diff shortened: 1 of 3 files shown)


test/publishToNpm.test.js
```diff
@@ -125,7 +125,7 @@ tap.test('Should stop action if package info retrieval fails', async t => {
   const { publishToNpmProxy, runSpawnStub } = setup()
 
   runSpawnStub
-    .withArgs('npm', ['view', 'fakeTestPkg@v5.1.3'])
+    .withArgs('npm', ['view', '--json'])
```
simoneb, line 128: is this test still accurate? shall we stop the action if package retrieval fails because of package not existing?

test/publishToNpm.test.js
```diff
@@ -155,3 +155,69 @@ tap.test('Should stop action if package info retrieval fails', async t => {
   ])
   t.pass('package is not published without otp code')
 })
+
+tap.test(
+  'Should stop action if package version info retrieval fails',
```
simoneb, line 160: this test has the same name as the previous one

src/utils/publishToNpm.js
```diff
@@ -2,6 +2,38 @@
...
+
+  // We need to check if the package was already published. This can happen if
+  // the action was already executed before, but it failed in its last step
+  // (GH release).
+  let packageName = null
+  try {
+    const packageInfo = await run('npm', ['view', '--json'])
+    packageName = packageInfo ? JSON.parse(packageInfo).name : null
+  } catch (error) {
+    if (!error?.message?.match(/code E404/)) {
```
simoneb, line 16: since we're asking to return json, don't we have access to json instead of doing a regexp against the full error message?

this is what happens when running this command on the terminal on a package that doesn't exist. you see there's both console output and json, but I'm not sure if we have both in the error object here. if we do have json, maybe it's better to use that instead

```sh
$ npm view --json
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/react-patterns-workshop - Not found
npm ERR! 404
npm ERR! 404  'react-patterns-workshop@latest' is not in this registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
{
  "error": {
    "code": "E404",
    "summary": "Not Found - GET https://registry.npmjs.org/react-patterns-workshop - Not found",
    "detail": "\n 'react-patterns-workshop@latest' is not in this registry.\nYou should bug the author to publish it (or use the name yourself!)\n\nNote that you can also install from a\ntarball, folder, http url, or git url."
  }
}
```

test/publishToNpm.test.js
ianlovell-nearform: isn't clear from the diff view, but the tests check for continue on a 404 from `npm view --json` and stopping for any other error

test/publishToNpm.test.js
ianlovell-nearform: the names are currently `package info retrieval` (`npm view --json`) and `package version info retrieval` (`npm view packagename@version`)

src/utils/publishToNpm.js
ianlovell-nearform: the stderr output for the command is currently part of the error message string only - https://github.com/nearform/optic-release-automation-action/blob/main/src/utils/runSpawn.js#L25-L29

simoneb (approved):

PR #120, opened 2022-05-23
Post comment on any issues that were closed by PRs included in release

xkal36:
Will add more tests when the approach is agreed upon.

Looks as follows:
Screenshot 2022-05-23 at 10 56 10

Approach is that, after the package is published on npm and released on github, we parse the release notes, get all PR numbers from that and for each PR, we fetch the description and check if there are any linked issues. After grabbing all the linked issue numbers, fort each one, we post a generic comment indicating that this is releases in whatever version and a link to the package on npm.


action.yml
```diff
@@ -73,6 +73,20 @@ runs:
...
+      run: |
+        echo "::set-output name=PACKAGE_VERSION::$(echo $(node -p "require('./package.json').version"))"
+      shell: 'bash'
+
+    - name: Get package name
+      id: get-package-name
+      if: ${{ github.event_name == 'pull_request' }}
+      run: |
+        echo "::set-output name=PACKAGE_NAME::$(echo $(node -p "require('./package.json').name"))"
+      shell: 'bash'
```
simoneb, line 88: a few comments on this:

- can't this be done in a single step?
- also, since we're setting the same variable earlier, I think there is a chance for refactoring it. plus, are we sure this is the easiest way to grab this information? I mean isn't it just simpler to do it in code than here?

action.yml
```diff
@@ -97,4 +111,4 @@ runs:
         github-token: ${{ inputs.github-token }}
         script: |
           const script = require('${{ github.action_path }}/dist/index.js')
-          await script({ github, context, inputs: ${{ toJSON(inputs) }} })
+          await script({ github, context, inputs: ${{ toJSON(inputs) }}, packageVersion: "${{ steps.get-package-version.outputs.PACKAGE_VERSION }}", packageName: "${{ steps.get-package-name.outputs.PACKAGE_NAME }}" })
```
simoneb, line 114: ultimately, and to summarize my earlier comments, isn't it just simpler to grab package version and name from within the code than here in the yaml?

src/index.js
```diff
@@ -4,13 +4,19 @@ const openPr = require('./openPr')
...
+  packageVersion,
+  packageName,
+}) {
   if (context.eventName === 'workflow_dispatch') {
     return openPr({ context, inputs, packageVersion })
   }
 
   if (context.eventName === 'pull_request') {
-    return release({ github, context, inputs })
+    return release({ github, context, inputs, packageVersion, packageName })
```
simoneb, line 19: back to my previous point, if we can assume that the code has access to the source code (it assumes we're doing a checkout in the yaml), then we can grab package name and version here in a much simpler way by loading the package json of the repo. beware of the paths though

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,94 @@
...
+          id
+          closingIssuesReferences(first: 100) {
+            nodes {
+              id
+              number
+            }
+          }
+        }
+      }
+    }
```
simoneb, line 21: is there a simpler way to do this with the rest API?

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,94 @@
...
+    `,
+      {
+        repoOwner,
+        repoName,
+        prNumber,
+      }
+    )
+    .then(queryResult => {
```
simoneb, line 29: let's use async-await please

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,94 @@
...
+  )
+
+  // Filter out the full change log numbers such as v1.0.1...v1.0.2
+  const prNumbers = potentialPrNumbers.filter(prNumber =>
+    prNumber.match(/^\d+$/)
+  )
+
+  const issueNumbersToNotify = (
+    await Promise.all(
+      prNumbers.map(async prNumber =>
```
simoneb, line 67: no need for async here if we're not awaiting the function we're invoking

src/release.js
```diff
@@ -119,6 +127,23 @@ module.exports = async function ({ github, context, inputs }) {
...
+    try {
+      // post a comment about release on npm to any linked issues in the
+      // any of the PRs in this release
+      await notifyIssues(
+        github,
+        release.body,
+        packageVersion, // npm version format ie no 'v' in front of version
+        owner,
+        repo,
+        release.html_url,
```
simoneb, line 139: this function is already getting many arguments. let's maybe provide the whole release in a single one instead of multiple. the release will probably also contain information about the repo and the owner I suppose

simoneb (commented):
Some early feedback about the implementation details. The approach overall is sound.

action.yml
denes-fekeshazy: The package info is now loaded in code as per suggestion.

src/index.js
denes-fekeshazy: The requested refactoring is done, the package info is not loaded in the notify issue util

src/utils/notifyIssues.js
denes-fekeshazy: I could not find a straight forward alternative in the API docs

src/utils/notifyIssues.js
denes-fekeshazy: Good spot, it has been removed.

src/release.js
denes-fekeshazy: The requested refactoring is done.

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,94 @@
...
+    )
+  ).flat()
+
+  const npmUrl = `https://www.npmjs.com/package/${packageName}/v/${npmVersion}`
+
+  const body = `🎉 This issue has been resolved in version ${npmVersion} 🎉 \n\n 
+  The release is available on: \n * [npm package (@latest dist-tag)](${npmUrl}) \n 
+  * [GitHub release](${releaseUrl}) \n\n Your **[optic](https://github.com/nearform/optic)** bot 📦🚀`
+
+  issueNumbersToNotify.forEach(async issueNumber => {
```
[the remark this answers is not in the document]
denes-fekeshazy: The requested refactoring is done. I had to use p-map 4.0.0, since from version 5 and onwards only ESM is supported.

action.yml
```diff
@@ -52,6 +52,11 @@ inputs:
     required: false
     type: boolean
     default: 'false'
+  notify-on-the-issue:
```
simoneb, line 55: let's call it notify-linked-issues

action.yml
```diff
@@ -52,6 +52,11 @@ inputs:
...
+    description: 'If the flag is set to "true", a comment will be posted on each issue that is linked to the pull requests for this release. The default setting is "true".'
```
simoneb, line 56:
```suggestion
    description: 'If the flag is set to "true", a comment will be posted on each issue that is linked to the pull requests included in each release. The default setting is "true".'
```

package.json
```diff
@@ -20,7 +20,9 @@
     "@actions/core": "^1.9.0",
     "@actions/github": "^5.0.3",
     "lodash.template": "^4.5.0",
+    "markdown-it": "^13.0.1",
     "node-fetch": "^2.6.6",
+    "p-map": "4.0.0",
```
simoneb, line 25: let's use semver anyway ^

src/release.js
```diff
@@ -119,6 +121,19 @@ module.exports = async function ({ github, context, inputs }) {
       inputs
     )
 
+    const shouldNotifyOnTheIssue = /true/i.test(inputs['notify-on-the-issue'])
```
simoneb, line 124: make sure the input name change reflects here as well of course

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,98 @@
...
+const pMap = require('p-map')
+
+const { logError, logWarning } = require('../log')
+const { getPrNumbersFromReleaseNotes } = require('./releaseNotes')
+
+async function getLinkedIssueNumbers({
+  octokit,
+  prNumber,
+  repoOwner,
+  repoName,
```
simoneb, line 13: for consistency, let's make these positional args

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,98 @@
...
+    )
+  ).flat()
+
+  const npmUrl = `https://www.npmjs.com/package/${packageName}/v/${packageVersion}`
+
+  const body = `🎉 This issue has been resolved in version ${packageVersion} 🎉 \n\n 
+  The release is available on: \n * [npm package (@latest dist-tag)](${npmUrl}) \n 
+  * [GitHub release](${releaseUrl}) \n\n Your **[optic](https://github.com/nearform/optic)** bot 📦🚀`
+
+  const mapper = async issueNumber => {
```
simoneb, line 86: let's give this a more meaningful name

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,98 @@
...
+  } catch (err) {
+    logWarning('Failed to get package info')
+    logError(err)
+  }
+
+  const { body: releaseNotes, html_url: releaseUrl } = release
+
+  const prNumbers = getPrNumbersFromReleaseNotes(releaseNotes)
+
+  const issueNumbersToNotify = (
```
simoneb, line 67: let's use p-map here as well

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,98 @@
...
+          repoName: repo,
+        })
+      )
+    )
+  ).flat()
+
+  const npmUrl = `https://www.npmjs.com/package/${packageName}/v/${packageVersion}`
+
+  const body = `🎉 This issue has been resolved in version ${packageVersion} 🎉 \n\n 
+  The release is available on: \n * [npm package (@latest dist-tag)](${npmUrl}) \n
```
simoneb, line 83: tag is not necessarily the latest, since we allow users to specify it as an input. either we take that into account, or we omit it from here

simoneb (commented):
See inline comments. Plus, we need more tests. Do we have coverage enabled? Let's increase the threshold of coverage and if not let's make sure we have it

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,98 @@
...
+
+  return linkedIssues.map(issue => issue.number)
+}
+
+async function notifyIssues(githubClient, owner, repo, release) {
+  let packageName
+  let packageVersion
+
+  try {
+    const packageJsonFile = fs.readFileSync('./package.json', 'utf8')
```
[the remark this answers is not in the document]
denes-fekeshazy: When I used `require` the contents of the file got evaluated at build time and pasted in the contents of the `package.json` from the `optic-release-automation-action` rather than the repo using the action. So I replaced it with the above to make sure that the `package.json` is read at runtime.

action.yml
denes-fekeshazy: The requested change is done.

action.yml
denes-fekeshazy: Change request applied.

package.json
denes-fekeshazy: The requested change is done.

src/release.js
denes-fekeshazy: This has also been updated to reflect the name change

src/utils/notifyIssues.js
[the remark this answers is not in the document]
denes-fekeshazy: At the time it was at the top of the `release.js` file as `const packageJsonFile = require('../package.json')`.
If I tried the path as `'./package.json'` then I got module not found on build time.

src/utils/notifyIssues.js
denes-fekeshazy: The requested change is done.

src/utils/notifyIssues.js
denes-fekeshazy: I have chose to omit the tag bit.

src/utils/notifyIssues.js
denes-fekeshazy: It has been renamed to `createCommentCallback`

test/releaseNotes.test.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+const tap = require('tap')
+
+const { getPrNumbersFromReleaseNotes } = require('../src/utils/releaseNotes')
+
+tap.test('Should return the correct PR numbers', async () => {
+  const testReleaseNotes =
+    "## What's Changed\n" +
```
simoneb, line 9: let's use a template string or even load from a file

.taprc
```diff
@@ -3,3 +3,7 @@ jsx: false
 flow: false
 check-coverage: false
 coverage: true
```
simoneb, line 5: I think all these options can removed as they have sensible defaults

README.md
```diff
@@ -186,6 +187,7 @@ jobs:
 | `build-command`| No       | An optional build commit to run after the version bump and before releasing the package |
 | `api-url`      | No       | GitHub App URL. You wouldn't need to set this unless you're deploying a custom GitHub app instead of [optic-release-automation](https://github.com/apps/optic-release-automation). <br /> (_Default: `https://optic-release-automation-ocrlhra4va-ue.a.run.app/`_) |
 | `sync-semver-tags`| No    | If you want to keep the major and minor versions git tags synced to the latest appropriate commit <br /> (_Default: `false`_)                                                                  |
+| `notify-on-the-issue`| No       | An optional flag to enable an automatic comment on all issues linked to the release so that people following those issues get notified of the fix being released. <br /> (_Default: `true`_)                                                                  |
```
simoneb, line 190: reflect name change, plus reword to avoid implying that an issue is a bug that needs to be fixed. it can be a feature or anything else

src/utils/notifyIssues.js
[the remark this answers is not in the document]
simoneb: yes of course if you put it at the top it's not the same as requiring it in the function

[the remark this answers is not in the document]
denes-fekeshazy: I have tried to replace `const packageJsonFile = fs.readFileSync('./package.json', 'utf8')` with `const packageJson = require('./package.json')` inside the function itself at the same line. Now I get `Error: Error: Cannot find module './package.json'` at runtime rather than build time. My understanding was that I should use the same path for both implementation. Any suggestions?

README.md
denes-fekeshazy: Updated copy:
```An optional flag to enable an automatic comment on all issues linked to each release so that people following those issues get notified of the code being released.```

Would this work or would you have an alternative suggestion?

File: test/releaseNotes.test.js (hunk 18 above)
Comment (denes-fekeshazy) [reply test/releaseNotes.test.js:9]: I have tried to turn that into a template string, but then the markdown parser perceived the whole release notes as a single `code_block` token, which resulted in incorrect output.

File: src/utils/notifyIssues.js (hunk 17 above)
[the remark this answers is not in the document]
Comment (simoneb) [reply src/utils/notifyIssues.js]: I'm not against reading the file but I'm confident there's a way to require the file. try github.workspace variable https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

File: test/releaseNotes.test.js (hunk 18 above)
Comment (simoneb) [reply test/releaseNotes.test.js:9]: that would be really weird because a template string is equivalent to concatenating string with new lines, except that the latter is syntactically uglier

File: README.md (hunk 20 above)
Comment (simoneb) [reply README.md]: this looks good

File: .taprc [hunk 21]
```diff
@@ -1,5 +1,7 @@
 ts: false
 jsx: false
 flow: false
-check-coverage: false
-coverage: true
```
simoneb, line 5: as I wrote before, all this can be removed

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,100 @@
...
+
+  if (!linkedIssues) {
+    return []
+  }
+
+  return linkedIssues.map(issue => issue.number)
+}
+
+async function notifyIssues(githubClient, owner, repo, release) {
+  console.log({
```
simoneb, line 44: avoid console log, use the action logger utilities

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,100 @@
...
+
+  const { body: releaseNotes, html_url: releaseUrl } = release
+
+  const prNumbers = getPrNumbersFromReleaseNotes(releaseNotes)
+
+  const getLinkedIssuesCallback = prNumber =>
+    getLinkedIssueNumbers(githubClient, parseInt(prNumber, 10), owner, repo)
+
+  const issueNumbersToNotify = (
+    await pMap(prNumbers, getLinkedIssuesCallback)
```
simoneb, line 79: just put the callback inline, it's simple enough to not deserve a dedicated named function

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,100 @@
...
+  const createCommentCallback = issueNumber => {
+    githubClient.rest.issues.createComment({
+      owner,
+      repo,
+      issue_number: issueNumber,
+      body,
+    })
+  }
+
+  await pMap(issueNumbersToNotify, createCommentCallback, { concurrency: 20 })
```
simoneb, line 97: inline the callback. reduce the concurrency

.taprc
denes-fekeshazy: Apart from the the threshold configs the rest is removed now.

src/utils/notifyIssues.js
denes-fekeshazy: the console log is for dev purposes and will be removed from the final version

.taprc
denes-fekeshazy: Only the threshold configs are present now.

src/utils/notifyIssues.js
denes-fekeshazy: The callback is moved inline as requested.

src/utils/notifyIssues.js
denes-fekeshazy: The callback is moved inline as requested and the concurrency is set to 10.

test/releaseNotes.test.js
denes-fekeshazy: My bad. The problem was that I have also removed the `\n` characters thinking that the markdown parser will pick up the line breaks from the template string.
Test string is now turned into a template string.

src/utils/notifyIssues.js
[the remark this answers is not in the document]
simoneb: which means it works, I assume?

src/utils/notifyIssues.js
```diff
@@ -0,0 +1,100 @@
...
+    release,
+  })
+
+  let packageName
+  let packageVersion
+
+  try {
+    // const packageJsonFile = fs.readFileSync('./package.json', 'utf8')
+    // const packageJson = JSON.parse(packageJsonFile)
+    const packageJson = require('./package.json')
```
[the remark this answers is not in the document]
denes-fekeshazy: The workspace value had to be fed in through the `action.yml` file and pass down the functions. With that in place I could use the proposed solution.

src/utils/notifyIssues.js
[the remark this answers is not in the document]
denes-fekeshazy: Yes, thank you for the pointers. The only thing left now is to refactor the relevant tests.

simoneb (approved):

PR #129, opened 2022-06-24
update optic link in the issue comment

### LINKED ISSUE #128: Tweak issue comment
- the comment mentions a package published to npm, but there's no package. This should be made conditional depending on whether we're publishing something to npm or not
- the link to optic should actually link to the action (meaning this repo), not the backend app

denes-fekeshazy:
Resolve #128 

- Replaced the URL for the `optic` bot to  point to https://github.com/nearform/optic-release-automation-action
- If the `npm-token` is missing from the inputs, then do not include the NPM package link in the comment placed on the issue

### DIFF
--- a/src/release.js
+++ b/src/release.js
@@ -75,10 +75,10 @@ module.exports = async function ({ github, context, inputs }) {
 
   const shouldRevertCommit = /true/i.test(inputs['revert-commit-after-failure'])
 
-  try {
-    const opticToken = inputs['optic-token']
-    const npmToken = inputs['npm-token']
+  const opticToken = inputs['optic-token']
+  const npmToken = inputs['npm-token']
 
+  try {
     if (npmToken) {
       await publishToNpm({ npmToken, opticToken, opticUrl, npmTag, version })
     } else {
@@ -129,7 +129,7 @@ module.exports = async function ({ github, context, inputs }) {
       try {
         // post a comment about release on npm to any linked issues in the
         // any of the PRs in this release
-        await notifyIssues(github, owner, repo, release)
+        await notifyIssues(github, npmToken, owner, repo, release)
       } catch (err) {
         logWarning('Failed to notify any/all issues')
         logError(err)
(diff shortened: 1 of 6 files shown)


src/release.js
```diff
@@ -129,7 +129,7 @@ module.exports = async function ({ github, context, inputs }) {
       try {
         // post a comment about release on npm to any linked issues in the
         // any of the PRs in this release
-        await notifyIssues(github, owner, repo, release)
+        await notifyIssues(github, npmToken, owner, repo, release)
```
simoneb, line 132: instead of passing around the npmToken, which is a secret, why don't we introduce a simple boolean which tells us if we need to publish to npm and pass that around instead? They're pretty much the same, but passing around something that's a secret piece of information just to detect if you need to do something or not doesn't sound good

denes-fekeshazy: The requested change is done for the `notifyIssues`, however the actual secret still needs to be passed on to the `publishToNpm` function.

simoneb: yes that's expected

src/utils/notifyIssues.js
```diff
@@ -39,7 +39,33 @@ async function getLinkedIssueNumbers(github, prNumber, repoOwner, repoName) {
...
+    return `🎉 This issue has been resolved in version ${packageVersion} 🎉 \n\n
+  The release is available on: \n * [npm package](${npmUrl}) 
+  \n * [GitHub release](${releaseUrl}) 
+  \n\n Your **[optic](https://github.com/nearform/optic-release-automation-action)** bot 📦🚀`
+  }
+
+  return `🎉 This issue has been resolved in version ${packageVersion} 🎉 \n\n
+  The release is available on: \n * [GitHub release](${releaseUrl}) 
+  \n\n Your **[optic](https://github.com/nearform/optic-release-automation-action)** bot 📦🚀`
+}
```
simoneb, line 60: I believe these strings don't need to contain \n, you can simply go to new lines. You may have to put more new lines than you expect for markdown to insert blank lines, for overall templates strings don't need to mix with explicit new lines.

simoneb (approved):

src/utils/notifyIssues.js
denes-fekeshazy: The requested changes are done.

PR #136, opened 2022-07-29
feat: add release artifact

davideroffo:
Closes https://github.com/nearform/optic-release-automation-action/issues/135

### What has been developed
- a new action input `release-artifact-build-folder`
- additional function `attachArtifact` for uploading an asset and attaching it to the release
- PR's template was updated with the addition of a new section for the Artifacts
- Implemented unit tests
- Updated README.md file with the three new inputs
- Added two new optional inputs: artifact-filename and artifact-label
- Replaced zip library with `adm-zip` capable of zipping single files and folders

### How I tested the implementation:

I created a new private repository with a script for `npm run build` where it builds a `dist` folder with some files.

After that, I've added the workflow for the release referencing this branch with the new action input:

```yaml
name: release

on:
  workflow_dispatch:
    inputs:
      semver:
        description: The semver to use
        required: true
        default: patch
        type: choice
        options:
          - patch
          - minor
          - major
  pull_request:
    types: [closed]

jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: nearform/optic-release-automation-action@feat/add-release-artifact
        with:
          github-token: ${{ secrets.github_token }}
          commit-message: 'chore: release {version}'
          build-command: |
            npm ci
            npm run build
          artifact-path: dist
          artifact-filename: customfilename.zip
          artifact-label: CUSTOMLABEL

```

### Screenshots:
- GitHub action run on a private repository for testing referencing this branch:

Screenshot 2022-07-29 at 16 06 11

- The pull request with the new release has a new section for the Artifacts:

Screenshot 2022-07-29 at 16 07 21

- RELEASE ASSET is now available in the Asset section for the release:

Screenshot 2022-07-29 at 16 09 40

- Custom filename and label for artifact, and download link in the PR body:

Screenshot 2022-08-01 at 14 01 01


action.yml
```diff
@@ -57,6 +57,9 @@ inputs:
     required: false
     type: boolean
     default: 'true'
+  release-artifact-build-folder:
```
simoneb, line 60: I'd keep this more generic and call it `artifact-path`. 

Also, since we're introducing this new feature and although we don't need it at the current time, I'd consider making it generic enough to support either folders or files. Then, in our code we can figure out if it's a folder or a file and in the first case zip the folder, in the second send the file as-is (or zip the file perhaps?)

dist/pr.tpl
```diff
@@ -21,6 +21,13 @@ Release author: @${author}
 - No major or minor tags will be updated as configured
 <% } %>
 
+#### Artifacts
+<% if (artifactAttached) { %>
+- An artifact will be attached to the release
+<% } else { %>
+- No artifacts will be attached to the release
+<% } %>
```
simoneb, line 29: since we're describing this in the PR, why not include the path of the artifact as well?

src/openPr.js
```diff
@@ -93,6 +95,25 @@ module.exports = async function ({ context, inputs, packageVersion }) {
...
+      const { id: releaseId } = draftRelease
+      try {
+        await attachArtifact(buildDir, releaseId, token)
+      } catch (err) {
+        logInfo(err.message)
+        core.setFailed(err.message)
+      }
+
+      logInfo('** Artifact attached! **')
+    }
```
simoneb, line 115: I wonder if it wouldn't be more appropriate to do this in the app instead of the action. I'm not sure, just checking

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,46 @@
...
+      repo,
+      release_id: releaseId,
+      data: fs.readFileSync(outFile),
+      name: outFile,
+      label: ASSET_LABEL,
+      headers,
+    })
+    return
+  } catch (err) {
+    console.log('error: ', err)
```
simoneb, line 41: remove

src/const.js
```diff
@@ -1,3 +1,5 @@
 'use strict'
 
 exports.PR_TITLE_PREFIX = '[OPTIC-RELEASE-AUTOMATION]'
+exports.ASSET_FILENAME = 'asset.zip'
+exports.ASSET_LABEL = 'RELEASE ASSET'
```
simoneb, line 5: let's make these optional action inputs with sensible defaults

src/openPr.js
```diff
@@ -93,6 +95,25 @@ module.exports = async function ({ context, inputs, packageVersion }) {
       },
       inputs
     )
+
+    // attach artifact
+    const buildDir = inputs['release-artifact-build-folder']
```
simoneb, line 100: I'm not sure this is the right place where to do this. The try/catch block where you positioned this code is for creating the PR, it doesn't look right to nest this code in there

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,46 @@
...
+
+const fs = require('fs')
+const { zip } = require('zip-a-folder')
+const github = require('@actions/github')
+const { ASSET_LABEL, ASSET_FILENAME } = require('../const')
+
+const attachArtifact = async (buildDir, releaseId, token) => {
+  const outFile = ASSET_FILENAME
+  try {
+    await zip(buildDir, outFile)
```
simoneb, line 11: have you thought about whether it's better to create a zip file or a tar archive?

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,46 @@
...
+  } catch (err) {
+    throw new Error(
+      'An error occurred while zipping the build folder: ' + err.message
+    )
+  }
+
+  // determine content-length for header to upload asset
+  const contentLength = filePath => fs.statSync(filePath).size
```
simoneb, line 19: when in an async context, always prefer the async variants of these operations

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  try {
+    const { owner, repo } = github.context.repo
+    const octokit = github.getOctokit(token)
+    await octokit.rest.repos.uploadReleaseAsset({
+      owner,
+      repo,
+      release_id: releaseId,
+      data: fs.readFileSync(outFile),
```
simoneb, line 34: async

simoneb (changes requested):

src/utils/attachArtifact.js
davideroffo: I was thinking of the zip file, but I'm happy to archive it as a tar file if needed. Do you see any benefit in doing a tar archive instead of a zip file? (probably one benefit should be that tar archives can be extracted on every filesystem without additional packages)

simoneb: no not really, let's just stick to whatever existing conventions github uses when adding artifacts to releases

README.md
```diff
@@ -188,6 +188,9 @@ jobs:
...
 | `sync-semver-tags`| No    | If you want to keep the major and minor versions git tags synced to the latest appropriate commit <br /> (_Default: `false`_)                                                                  |
 | `notify-linked-issues`| No       | An optional flag to enable an automatic comment on all issues linked to each release so that people following those issues get notified of the code being released. <br /> (_Default: `true`_)                                                                  |
+| `artifact-path`| No       | If this property is populated, the item (folder or file) specified here will be used as the artifact to zip and attach to the release.                                                                  |
+| `artifact-filename`| No       | The artifact file name. <br /> (_Default: `asset.zip`_)                                                                  |
+| `artifact-label`| No       | The artifact label that will be shown in the pull request and in the release. <br /> (_Default: `RELEASE ASSET`_)                                                                  |
```
simoneb, line 193: what is the difference between the file name and the label? I think we should be using more sensible defaults. Instead of a fixed (although customizable) value, why don't we default them to the name of the file/folder pointed to by `artifact-path`?

action.yml
```diff
@@ -57,6 +57,17 @@ inputs:
...
+    description: 'If this property is populated, the item (folder or file) specified here will be used as the artifact to zip and attach to the release.'
+    required: false
+  artifact-filename:
+    description: 'The artifact file name.'
+    required: false
+    default: 'asset.zip'
+  artifact-label:
+    description: 'The artifact label that will be shown in the pull request and in the release.'
+    required: false
+    default: 'RELEASE ASSET'
```
simoneb, line 70: clearly if we make them depend on the value of artifact-path, then we can't provide default values here (which is fine)

src/openPr.js
```diff
@@ -73,11 +78,44 @@ module.exports = async function ({ context, inputs, packageVersion }) {
...
+        releaseId,
+        token
+      ))
+    } catch (err) {
+      logInfo(err.message)
+      core.setFailed(err.message)
+    }
+
+    logInfo('** Artifact attached! **')
+  }
```
simoneb, line 111: let's avoid mutable state. iow, let's avoid let artifact, then the weird ;({ ... }) = await syntax. also, let's extract this in a function please, this function is becoming too complex

src/pr.tpl
```diff
@@ -21,6 +21,13 @@ Release author: @${author}
 - No major or minor tags will be updated as configured
 <% } %>
 
+#### Artifacts
+<% if (artifact.isPresent) { %>
```
simoneb, line 25: why this logic? can't we just do

```suggestion
<% if (artifact) { %>
```

src/pr.tpl
```diff
@@ -21,6 +21,13 @@ Release author: @${author}
...
+- An artifact will be attached to the release -> [${artifact.label}](${artifact.url})
+<% } else { %>
+- No artifacts will be attached to the release
```
simoneb, line 28: we don't need to print this if the user has not specified an artifact

src/utils/archiver.js
```diff
@@ -0,0 +1,35 @@
...
+const archiveItem = async (path, out) => {
+  let isDirectory = false
+  try {
+    const stat = await lstat(path)
+    isDirectory = stat.isDirectory()
+  } catch (err) {
+    throw new Error(
+      'An error occurred while checking if file or directory: ' + err.message
+    )
+  }
```
simoneb, line 15: avoid this mutable state pattern. extract a function and you don't need this pattern as you can simply return the value

src/utils/archiver.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const zip = new AdmZip()
+  try {
+    if (isDirectory) {
+      await zip.addLocalFolderPromise(path)
+    } else {
+      zip.addLocalFile(path)
```
simoneb, line 22: what a silly API this library has, mixed async/sync, very confusing!

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,57 @@
...
+      repo,
+      release_id: releaseId,
+      data,
+      name: filename,
+      label,
+      headers,
+    })
+
+    if (!postAssetResponse.data) {
+      throw new Error('POST asset response data not available')
```
simoneb, line 39: is this the right way to catch errors?

src/utils/archiver.js
```diff
@@ -0,0 +1,35 @@
...
+    await zip.writeZipPromise(out)
+  } catch (err) {
+    throw new Error(
+      'An error occurred while zipping the build folder: ' + err.message
+    )
+  }
+}
+
+module.exports = {
+  archiveItem,
```
simoneb, line 34: let's use a consistent way to export stuff from modules

src/utils/attachArtifact.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const { browser_download_url: url, label: assetLabel } =
+      postAssetResponse.data
+
+    return {
+      artifact: {
+        isPresent: true,
```
simoneb, line 47: this ispresent is redundant

test/archiver.test.js
```diff
@@ -0,0 +1,21 @@
...
+      lstat: () => ({
+        isDirectory: () => {
+          throw Error()
+        },
+      }),
+    },
+  })
+
+  try {
+    await archiverModule.archiveItem()
```
simoneb, line 17: if this doesn't fail, this test will pass

test/attachArtifact.test.js
```diff
@@ -0,0 +1,193 @@
...
+          request: async () => ({
+            data: { browser_download_url: 'https://example.com' },
+          }),
+        }),
+      },
+    })
+
+    const { artifactPath, artifactFilename, artifactLabel, releaseId, token } =
+      DEFAULT_INPUT_DATA
+    tap.doesNotThrow(async () => {
```
simoneb, line 47: there are better methods to check promise resolution/rejection

test/attachArtifact.test.js
```diff
@@ -0,0 +1,193 @@
...
+  }
+)
+
+tap.test(
+  'attachArtifact throws an error if build folder not found',
+  async () => {
+    const error = new Error('file not found')
+    const attachArtifactModule = tap.mock('../src/utils/attachArtifact.js', {
+      '../src/utils/archiver.js': {
+        archiveItem: () => {
```
simoneb, line 65: isn't this function supposed to be async?

test/attachArtifact.test.js
```diff
@@ -0,0 +1,193 @@
...
+    try {
+      await attachArtifactModule.attachArtifact(
+        artifactPath,
+        artifactFilename,
+        artifactLabel,
+        releaseId,
+        token
+      )
+    } catch (err) {
+      tap.equal(err.message, expectedError.message)
```
simoneb, line 85: if the code before does not throw, this test will pass

test/attachArtifact.test.js
```diff
@@ -0,0 +1,193 @@
...
```
simoneb, line 139: same as previous comments, this test will succeed if no error is thrown

simoneb (changes requested):

test/attachArtifact.test.js
davideroffo: could you please name a few? I found only this one in the node-tap documentation. Thank you

simoneb: https://node-tap.org/docs/api/asserts/#trejectspromise--fn-expectederror-message-extra

README.md
```diff
@@ -188,6 +188,8 @@ jobs:
 | `api-url`      | No       | GitHub App URL. You wouldn't need to set this unless you're deploying a custom GitHub app instead of [optic-release-automation](https://github.com/apps/optic-release-automation). <br /> (_Default: `https://optic-release-automation-ocrlhra4va-ue.a.run.app/`_) |
 | `sync-semver-tags`| No    | If you want to keep the major and minor versions git tags synced to the latest appropriate commit <br /> (_Default: `false`_)                                                                  |
 | `notify-linked-issues`| No       | An optional flag to enable an automatic comment on all issues linked to each release so that people following those issues get notified of the code being released. <br /> (_Default: `true`_)                                                                  |
+| `artifact-path`| No       | If this property is populated, the item (folder or file) specified here will be used as the artifact to zip and attach to the release.                                                                  |
+| `artifact-filename`| No       | The artifact file name with the specific extension.                                                                 |
```
simoneb, line 192: it's not clear from here what this input does

action.yml
```diff
@@ -57,6 +57,12 @@ inputs:
     required: false
     type: boolean
     default: 'true'
+  artifact-path:
```
simoneb, line 60: what if this is already a zip?

src/openPr.js
```diff
@@ -73,11 +78,37 @@ module.exports = async function ({ context, inputs, packageVersion }) {
 
   logInfo(`New version ${newVersion}`)
 
+  // attach artifact
+  const artifactPath = inputs['artifact-path']
```
simoneb, line 82: extract all this logic to a function please

test/artifact.test.js
```diff
@@ -0,0 +1,157 @@
...
+  'attach artifact throws an error if an error occurres during the asset upload',
+  async () => {
+    const artifactModule = tap.mock('../src/utils/artifact.js', {
+      '../src/utils/archiver.js': {
+        archiveItem: async () => null,
+      },
+      'fs/promises': {
+        stat: async () => 100,
+        lstat: async () => ({ isDirectory: () => true }),
+        readFile: async () => Buffer.from([0x62, 0x75, 0x66, 0x66, 0x65, 0x72]),
```
simoneb, line 88: for readability, here and elsewhere

```suggestion
        readFile: async () => Buffer.from('hello!'),
```

action.yml
davideroffo: I added a check, if the file is already a zip file the procedure won't zip it.

README.md
davideroffo: I tried to change the wording in order to be more clear

src/openPr.js
```diff
@@ -34,11 +38,34 @@ const getPRBody = (template, { newVersion, draftRelease, inputs, author }) => {
...
+    artifact,
     syncTags: /true/i.test(inputs['sync-semver-tags']),
     author,
   })
 }
 
+const addArtifact = async (inputs, releaseId) => {
+  const artifactPath = inputs['artifact-path']
+
+  if (artifactPath) {
```
simoneb, line 50: leave to the calling code the decision to invoke this function or not

src/openPr.js
```diff
@@ -73,11 +100,20 @@ module.exports = async function ({ context, inputs, packageVersion }) {
 
   logInfo(`New version ${newVersion}`)
 
+  let artifact
+  try {
+    artifact = await addArtifact(inputs, draftRelease.id)
+  } catch (err) {
+    logInfo(err.message)
+    core.setFailed(err.message)
+  }
```
simoneb, line 109:
```suggestion
  const artifact = inputs['artifact-path'] && addArtifact(inputs['artifact-path'], draftRelease.id)
```

src/utils/artifact.js
```diff
@@ -0,0 +1,62 @@
...
+const { archiveItem } = require('./archiver')
+const { ZIP_EXTENSION } = require('../const')
+
+const attach = async (path, filename, releaseId, token) => {
+  if (!path.endsWith(ZIP_EXTENSION)) {
+    try {
+      await archiveItem(path, filename)
+    } catch (err) {
+      throw new Error(err.message)
+    }
```
simoneb, line 15:
```suggestion
      await archiveItem(path, filename)
```

test/artifact.test.js
```diff
@@ -0,0 +1,157 @@
...
+)
+
+tap.test(
+  'attach artifact throws an error if build folder not found',
+  async () => {
+    const artifactModule = tap.mock('../src/utils/artifact.js', {
+      '../src/utils/archiver.js': {
+        archiveItem: async () => {
+          throw new Error('file not found')
+        },
```
simoneb, line 65: as an alternative: Promise.reject(new Error('...'))

test/bump.test.js
```diff
@@ -235,6 +247,8 @@ tap.test(
             '- No major or minor tags will be updated as configured\n' +
             '\n' +
             '\n' +
+            '\n' +
+            '\n' +
```
simoneb, line 251: can we avoid adding blank lines if there is no artifact?

simoneb (approved):

davideroffo:
All the request changes have been addressed. Merging ...

PR #151, opened 2022-09-06
fix: Pull Request reopened delete the draft release

CristiTeo:
Added the approach to delete the draft release when the pull request is reopened.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -26347,7 +26347,7 @@ module.exports = async function ({ github, context, inputs }) {
   const repo = context.repo.repo
 
   if (
-    context.payload.action !== 'closed' ||
+    !['closed', 'reopened'].includes(context.payload.action) ||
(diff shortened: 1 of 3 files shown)


simoneb:
please link the issue

CristiTeo:
From what I understand in the example that we have the PR was closed and after that reopened.  
In this case, we have 2 new release branches, and when they were merged into the master caused the duplicated comments. 
For that, we had a TODO comment exactly for this case, so I implemented an approach for this case.

simoneb:
> From what I understand in the example that we have the PR was closed and after that reopened. In this case, we have 2 new release branches, and when they were merged into the master caused the duplicated comments. For that, we had a TODO comment exactly for this case, so I implemented an approach for this case.

That is not what happened. The PR was not reopened, a new one was created.

PR #168, opened 2022-10-24
chore: replace set-output

brunopascottini:
GitHub Actions: Deprecating save-state and set-output commands.

GitHub actions using the mentioned commands will start to fail from 1st of June of 2023.
[Reference](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -83,7 +83,7 @@ runs:
       if: ${{ github.event_name == 'workflow_dispatch' }}
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
-        echo "::set-output name=PACKAGE_VERSION::$(echo $(node -p "require('./package.json').version"))"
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -83,7 +83,7 @@ runs:
       if: ${{ github.event_name == 'workflow_dispatch' }}
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
-        echo "::set-output name=PACKAGE_VERSION::$(echo $(node -p "require('./package.json').version"))"
+        echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
```
simoneb, line 86: can you check if this command still works properly? I realize it's very similar to the previous one, but the nested double quotes are a little suspicious

brunopascottini: Running the shell command in this same repository gets "PACKAGE_VERSION=3.2.4", which is correct

simoneb (approved):

PR #171, opened 2022-11-08
feat: use branchName as release target

### LINKED ISSUE #167: Sanity check: is the action releasing the right version or the latest commit?
When a PR created by the action gets merged, which version of the repo is released exactly? 

E.g.

- PR opened in a branch with commit 1
- commits 2, 3, 4 are done on master
- PR merged, causing merge commit 5

Theoretically, the semantics of the action should lead to the result of merging commit 1 to the master branch to be released, but I guess that what's happening now is that all commits in between will end up in the release.

We need to figure out if there's a way to make this work according to correct semantics.

guilhermelimak:
Set the release branch name as target to avoid merging the changes commited to the main branch after the release PR is created to be included in the release.

References #167

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ jobs:
           npm-tag: ${{ github.event.inputs.tag }}
 ```
 
-> Not all symbols that can be used in GitHub usernames are valid in secret names. One such example is the hyphen symbol (`-`). In such cases, this approach will not work. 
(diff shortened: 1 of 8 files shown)


guilhermelimak:
@simoneb This is ready for review

simoneb (approved):

PR #178, opened 2022-11-17
Improve error handling for non existent releases

guilhermelimak:


simoneb (commented):
This looks good to me but apart from error handling it doesn't seem to introduce any functional changes from the original implementation. I don't mind merging it, but did we figure out what was the cause of the issue?

src/openPr.js
```diff
@@ -53,6 +53,33 @@ const addArtifact = async (inputs, releaseId) => {
...
+        },
+      },
+      inputs
+    )
+
+    logInfo(`Draft release created successfully`)
+
+    return draftRelease
+  } catch (err) {
+    core.setFailed(`Unable to create draft release: ${err.message}`)
```
simoneb, line 79: are we sure this stops the execution immediately? because otherwise we're catching the error and execution continues, even if just for a short time

simoneb (approved):
I'll approve to avoid blocking. Needless to say it would be ideal to figure out the cause of the issue

simoneb (commented):
let's repurpose this PR in terms of failing consistently when something unexpected happens:

1. the draft release could not be created
2. the draft release could not be found when trying to turn it into a published release

guilhermelimak:
@simoneb I changed some small stuff and fixed the tests. After a quick review it should be ready to merge.

It should now fail in both cases:
> 1. the draft release could not be created
> 2. the draft release could not be found when trying to turn it into a published release

src/openPr.js
```diff
@@ -75,19 +103,12 @@ module.exports = async function ({ context, inputs, packageVersion }) {
...
-        version: newVersion,
-        target: releaseCommitHash,
-      },
-    },
-    inputs
-  )
+  // Halt action here if release draft could not be created for some reason.
+  if (!draftRelease) {
+    throw new Error('Failed to create draft release. Aborting.')
+  }
```
simoneb, line 111: in theory this check should be unnecessary. having it means that we're not sure if the resolved value of callApi may be falsy, which in my opinion would signal a bug elsewhere. if we do want to preserve this check, I'd move it inside the createDraftRelease function though, for consistency, so we don't have to do fancy error handling here

src/release.js
```diff
@@ -42,6 +42,24 @@ module.exports = async function ({ github, context, inputs }) {
...
   const { opticUrl, npmTag, version, id } = releaseMeta
 
+  try {
+    const { data: draftRelease } = await github.rest.repos.getRelease({
+      owner,
+      repo,
+      release_id: id,
+    })
+
+    if (!draftRelease || !draftRelease.draft) {
```
simoneb, line 52: we should not check if the release is draft, because this would make this non-idempotent, meaning that we can't run it twice if anything breaks. it is enough to check if a release is there, if it is, we're good, even if it's a published release (remember we _explicitly_ talked about this)

test/release.test.js
```diff
@@ -570,3 +572,43 @@ tap.test(
     sinon.assert.notCalled(stubs.notifyIssuesStub)
   }
 )
+
+tap.test('Should fail when release is not found', async () => {
```
simoneb, line 576: don't we need tests also for the PR part of this action, to cover the error cases you added in this PR?

simoneb (changes requested):

test/release.test.js
guilhermelimak: Yep, added a new test for that case

src/openPr.js
guilhermelimak: I agree it shouldn't be necessary but since fetch will only throw on network failure and not if the response is 4xx or 5xx I thought it was a good idea to add that just to make sure we're not missing anything but I removed it now. Maybe we can improve `callApi` later to handle errors better, since now it checks the reponse status but if it's not 200 we just log a warning telling the user to install the github app (I'm not even sure that makes 100% sense since (at least I think) we could have something different causing a server error, not just that the app is not installed)

simoneb (approved):

PR #181, opened 2022-11-21
Fix: external linked issues notification

### LINKED ISSUE #179: Handle external linked issues
When the action does a release it by default tries to comment on the linked issues of each PR included in the release, mentioning in which release the issue was addressed.

We came across a case whereby the linked issue was in a different repo, which failed the action. See [here](https://github.com/nearform/fastify-constraints/actions/runs/3496077957/jobs/5853571358). 

The action needs to handle this more correctly and post the comment on the issue in a "best effort" way. It should still try, but it shouldn't fail the execution if it cannot comment. Also note that I'm not even sure this case should have failed, it's just that the action tried to post the comment on an issue that doesn't exist because it was searching for that issue in the same repo, whereas it was an external repo.

### LINKED ISSUE #180: js error in release workflow
See https://github.com/nearform/fast-jwt/actions/runs/3496610299/jobs/5854718164

sameer-coder:
This PR fixes the issues #179 and #180

The action was failing when trying to create a comment on an issue linked to an external repository. That error is handled correctly now and the action no longer fails.

closes #179 
closes #180


src/utils/releaseNotes.js
```diff
@@ -3,7 +3,7 @@
 const md = require('markdown-it')()
 
 function getPrNumbersFromReleaseNotes(releaseNotes) {
-  const parsedReleaseNotes = md.parse(releaseNotes)
+  const parsedReleaseNotes = md.parse(releaseNotes, {})
```
simoneb, line 6: why this change?

simoneb (approved):

PR #186, opened 2022-11-23
feat: add preliminary support for monorepos

guilhermelimak:


simoneb (commented):
so far so good, it looks like a reasonable approach

guilhermelimak:
@simoneb It's using @actions/exec now and should be ready to test.

src/index.js
```diff
@@ -15,3 +16,5 @@ module.exports = async function ({ github, context, inputs, packageVersion }) {
 
   logError('Unsupported event')
 }
+
+module.exports.build = build
```
simoneb, line 20: ok this is not super, in the sense that I wonder if along this change we shall consider also splitting the other two in two functions instead of having the multiplexing logic in the main export. but ok, we can keep it as it is for now

src/pr.tpl
```diff
@@ -3,7 +3,9 @@
 This **draft** PR is opened by Github action [optic-release-automation-action](https://github.com/nearform/optic-release-automation-action).
 
 A new **draft** GitHub release [${releaseMeta.version}](${draftRelease.html_url}) has been created.
-
+<% if (releaseMeta.monorepoPackage) { %>
```
simoneb, line 6: instead of just adding this, I think we should consider rewording the existing text of this template instead. If you think it's too early to make this change, ok, but let's track this as an issue.

src/release.js
```diff
@@ -28,19 +29,14 @@ module.exports = async function ({ github, context, inputs }) {
...
-  let releaseMeta
+  let releaseMetadata
   try {
-    releaseMeta = JSON.parse(
-      pr.body.substring(
-        pr.body.indexOf('<release-meta>') + 14,
-        pr.body.lastIndexOf('</release-meta>')
-      )
-    )
+    releaseMetadata = parseReleaseMetadata(pr)
```
simoneb, line 34: since you're moving this logic into a function, there's no need to keep error handling here, let's just do it in the function and just let an error bubble up in case it breaks, so the conditional logic here becomes simpler

src/build.js
```diff
@@ -0,0 +1,38 @@
...
+const core = require('@actions/core')
+const { exec } = require('@actions/exec')
+const parseReleaseMetadata = require('./utils/parseReleaseMetadata')
+const { logError } = require('./log')
+
+module.exports = async function ({ context, inputs }) {
+  let releaseMetadata
+
+  try {
+    releaseMetadata = parseReleaseMetadata(context.payload.pull_request)
```
simoneb, line 12: just a question. how do we know that context.payload.pull_request is populated?

simoneb (commented):
Overall looks good, although this is still missing the release part which we can do later but which is necessary before we roll this out fully. A few comments though.

action.yml
```diff
@@ -69,6 +69,13 @@ inputs:
     description: 'A prefix to apply to the version number, which reflects in the tag and GitHub release names'
     required: true
     default: 'v'
+  monorepo-package:
+    description: 'A specific monorepo package to be deployed. The package should have the full path from the root of the project E.g. `packages/react-app`'
```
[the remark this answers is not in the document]
guilhermelimak: Yep this was a mistake. At first I thought it would be a good idea for them to be defined together but that quickly proved to be a mistake and I forgot to update the docs later.

src/index.js
guilhermelimak: I removed the multiplexing logic and changed the other exports. It was a pretty simple change so I think it won't cause any issues but I can just revert it if you think it's a bad idea

src/utils/parseReleaseMetadata.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+module.exports = function getReleaseMetadata(pr) {
+  return JSON.parse(
```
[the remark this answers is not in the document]
guilhermelimak: I just added a catch to to throw with a message that says the error was caused when trying to parse the release metadata to make it a bit more clear.

README.md
```diff
@@ -168,6 +168,25 @@ jobs:
...
+  workflow_dispatch:
+#...
+    monorepo-package:
+      description: "The package to be released"
+      required: true
+      type: choice
+      options:
+        - packages/express-server
+        - packages/react-app
+        - packages/shared-data
```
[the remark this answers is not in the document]
guilhermelimak: There's no way to use it without monorepo-root, it will always default to packages if the user doesn't supply it.

src/build.js
guilhermelimak: I think it will always be there since this step will only run for pull requests as defined herr but I'm not entirely sure.
https://github.com/nearform/optic-release-automation-action/blob/e60e3dd3b4f71f1d7ee1746600e634c6de58f49d/action.yml#L102

src/pr.tpl
guilhermelimak: I have created [this issue](https://github.com/nearform/optic-release-automation-action/issues/192) to track the template improvements

README.md
```diff
@@ -187,7 +219,10 @@ jobs:
...
 | `artifact-path`| No       | Set this input to the distribution folder or file you want to add as the main asset for your release. It will be downloadable from the release page and a preview of it will be available in the pull request.                                                                  |
-| `version-prefix`       | No       | A prefix to apply to the version number, which reflects in the tag and GitHub release names. <br /> (_Default: 'v'_)                                                                                                                                                                                                                                         |
+| `version-prefix`       | No       | A prefix to apply to the version number, which reflects in the tag and GitHub release names. <br /> (_Default: 'v'_)|
+| `monorepo-package`     | No       | The monorepo package folder to be released without path |
```
simoneb, line 223:
```suggestion
| `monorepo-package`     | No       | The name of the monorepo package folder to be released, relative to `monorepo-root` |
```

src/build.js
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    await exec('node', ['-v'], options)
+    await exec('npm', ['-v'], options)
+
+    for (const [command, ...args] of buildCommands) {
+      await exec(command, args, options)
+    }
+  } catch (err) {
+    core.setFailed(`Error when building package: ${err.message}`)
```
simoneb, line 38: let's include the name of the package in the error message

src/build.js
```diff
@@ -0,0 +1,40 @@
...
+  return {
+    monorepoPackage: inputs['monorepo-package'],
+    monorepoRoot: inputs['monorepo-root'],
+  }
+}
+
+module.exports = async function ({ context, inputs }) {
+  const { monorepoPackage, monorepoRoot } = getMonorepoData({ context, inputs })
+
+  const buildCommands = inputs['build-command']
```
simoneb, line 21: I see what you're doing here. It looks reasonable, but probably not very robust. 

Very simple example: what if the user puts an additional new line in the command? I haven't tried but I suspect it'll try to execute an empty command, thereby failing. This one may be easy to fix, but I wonder what other cases we may come across. Again this is because we're pretending to be a shell, which we aren't

action.yml
```diff
@@ -91,27 +98,29 @@ runs:
...
     - name: Release the package
       uses: actions/github-script@v6
       if: ${{ github.event_name == 'pull_request' }}
       with:
         github-token: ${{ inputs.github-token }}
         script: |
-          const script = require('${{ github.action_path }}/dist/index.js')
-          await script({ github, context, inputs: ${{ toJSON(inputs) }} })
+          const { release } = require('${{ github.action_path }}/dist/index.js')
+          await release({ github, context, inputs: ${{ toJSON(inputs) }} })
```
simoneb, line 126: missing a new line here

simoneb (commented):
looking better, still a few comments

src/build.js
guilhermelimak: Yes, I just fixed the case you mentioned, but I'm really not sure about how to avoid doing this in javascript. I tried one more time to go back to bash but we do need to run it when closing the PR to publish the npm package. 

One thing we can do is instead of using the release body to store metadata we just save it to a .json file and use [upload-artifact](https://github.com/actions/upload-artifact) and [download-artifact](https://github.com/actions/download-artifact) actions to store and retrieve it between runs. This would not be a simple change tho so I think it should be done in another PR if you think that's a good idea.

simoneb:
As discussed, let's hold on merging this and start using this branch from a couple repos having a build-command configured, then let's wait until we do a release of those repos and if all good, we merge this and do a new release.

src/build.js
simoneb: no need to go that far as using artifacts

guilhermelimak:
I pinned the action to our branch in the following repositories:
- https://github.com/nearform/github-snooze-chrome-extension/
- https://github.com/nearform/temporal_tables
- https://github.com/nearform/optic-expo/

After we're done with testing those should be reverted back to the previous version

README.md
```diff
@@ -168,6 +168,38 @@ jobs:
             npm run build
 ```
 
+## Releasing monorepo packages
+
+For use in a monorepo this action expects two new inputs `monorepo-package` and `monorepo-root`, they will be used to run command and publish from the package folder instead of the project root.
```
penx, line 173: what is the "project root" here? I don't see this as a configurable option, so at the moment I assume this means "repository root"?

penx: if you had a single package, but it lived in a subdirectory of your repository, could monorepo-root be used without monorepo-package?

Perhaps just a "root" is needed, or maybe "working-directory", "workingDir"?

penx: Given a repository of the structure

```
yarn.lock
|-packages
  |-my-package-1
  |-my-package-2
```

I would expect the name "monorepo-root" to mean where your lock file is.

But I think you are using it here to refer to the packages directory.

I wonder if packages could include paths, so for the above you would have

```yml
root: .
packages:
  - packages/my-package-1
  - packages/my-package-2
```

And though possibly uncommon would support a repository that looked like:


``
my-java-project
|-Jenkinsfile.jank
my-node-project
|-yarn.lock
  |-packages
    |-my-package-1
    |-my-package-2
```

with

```yml
root: my-node-project
packages:
  - packages/my-package-1
  - packages/my-package-2
```

test/build.test.js
```diff
@@ -0,0 +1,231 @@
...
+    data.inputs['monorepo-package'] = 'react-app'
+    data.inputs['monorepo-root'] = 'packages'
+    data.inputs['build-command'] = 'npm  install\n\nnpm   build'
+
+    data.context.payload.pull_request.body =
+      '<!--\n' +
+      '<release-meta>{"id":54503465,"version":"v5.1.3","npmTag":"latest", "monorepoPackage": "another-package", "monorepoRoot": "packages"}</release-meta>\n' +
+      '-->'
+
+    const cwd = `${data.inputs['monorepo-root']}/${data.inputs['monorepo-package']}`
```
penx, line 220: as far as I can see
- monorepo-root is only used here
- monorepo-package could contain a slash

...which kind of makes monorepo-root redundant in its current form? (other than my point above about supporting monorepos in a sub folder)

src/build.js
```diff
@@ -0,0 +1,53 @@
...
+      .filter(cmd => !!cmd)
+      .map(cmd =>
+        cmd
+          .trim()
+          .split(' ')
+          .filter(c => !!c)
+      )
+
+    const options = {
+      cwd: monorepoPackage ? `${monorepoRoot}/${monorepoPackage}` : '.',
```
penx, line 41: as far as I can see
- monorepoRoot is only used for generating directory names
- monorepoPackage could contain a slash

...which kind of makes monorepo-root redundant in its current form? (other than my point above about supporting monorepos in a sub folder)

src/release.js
```diff
@@ -61,7 +51,9 @@ module.exports = async function ({ github, context, inputs }) {
   }
 
   const run = runSpawn()
-  const branchName = `release/${version}`
+  const branchName = monorepoPackage
+    ? `release/${monorepoPackage}-${version}`
```
penx, line 55: could you `require('./package.json').name` here to get the specific package name?

guilhermelimak: I'm not sure this would be a good idea since the package name could include a namespace such as @babel/preset-env

README.md
guilhermelimak: > if you had a single package, but it lived in a subdirectory of your repository, could monorepo-root be used without monorepo-package?
> 
> Perhaps just a "root" is needed, or maybe "working-directory", "workingDir"?

I agree it could be useful but I think it would be better suited for another PR since it isn't related to monorepos

guilhermelimak:
Pin this branch in those 3 repos again for testing:

https://github.com/nearform/github-snooze-chrome-extension/pull/487
https://github.com/nearform/temporal_tables/pull/47
https://github.com/nearform/optic-expo/pull/995

davideroffo (changes requested):

src/openPr.js
```diff
@@ -150,12 +171,14 @@ module.exports = async function ({ context, inputs, packageVersion }) {
   } catch (err) {
     let message = `Unable to create the pull request ${err.message}`
     try {
-      await execWithOutput('git', ['push', 'origin', '--delete', branchName])
+      const exec = execWithOutput()
+      await exec('git', ['push', 'origin', '--delete', branchName])
```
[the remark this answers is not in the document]
guilhermelimak: it's not really for readability but to avoid having to pass the cwd parameter for all function calls. Since most of the time when using this function all calls will be related and run on the same directory it makes sense for me to pass it once and then it will be set for all subsequent calls.

simoneb:
Closing this as it's been stale for some time and apparently we're not in such a rush to get monorepo support. Thanks nonetheless @guilhermelimak for all your work on this

PR #196, opened 2022-12-05
Fix: Truncate PR body if exceeds max size

### LINKED ISSUE #194: No error on PR creation failure
When the PR creation fails the action still reports successful execution.

### LINKED ISSUE #197: Pull request creation error - body is too long
If the release notes are too long(more than 65536 characters) then PR creation fails.

sameer-coder:
This PR fixes 2 issues:
- Truncates PR body if it exceeds the 65536 characters limitation set by GitHub
- Sets the action execution to failed if PR creation process fails

Closes #194 
Closes #197

### DIFF
--- a/src/openPr.js
+++ b/src/openPr.js
@@ -3,9 +3,7 @@
 const fs = require('fs')
 const path = require('path')
 const _template = require('lodash.template')
-const semver = require('semver')
 const core = require('@actions/core')
-const _truncate = require('lodash.truncate')
 
 const { PR_TITLE_PREFIX } = require('./const')
 const { runSpawn } = require('./utils/runSpawn')
@@ -13,51 +11,9 @@ const { callApi } = require('./utils/callApi')
 const transformCommitMessage = require('./utils/commitMessage')
 const { logInfo } = require('./log')
 const { attach } = require('./utils/artifact')
+const { getPRBody } = require('./utils/releaseNotes')
 
 const tpl = fs.readFileSync(path.join(__dirname, 'pr.tpl'), 'utf8')
-
-const MAX_PR_BODY_SIZE_LIMITATION = 65536
-const PR_BODY_TRUNCATE_SIZE = 60000
-
-const getPRBody = (
-  template,
-  { newVersion, draftRelease, inputs, author, artifact }
-) => {
-  const tagsToBeUpdated = []
-  const { major, minor } = semver.parse(newVersion)
-
-  if (major !== 0) tagsToBeUpdated.push(`v${major}`)
-  if (minor !== 0) tagsToBeUpdated.push(`v${major}.${minor}`)
-
-  // Should strictly contain only non-sensitive data
-  const releaseMeta = {
-    id: draftRelease.id,
-    version: newVersion,
-    npmTag: inputs['npm-tag'],
-    opticUrl: inputs['optic-url'],
-  }
-
-  const prBody = template({
-    releaseMeta,
-    draftRelease,
-    tagsToUpdate: tagsToBeUpdated.join(', '),
-    npmPublish: !!inputs['npm-token'],
-    artifact,
-    syncTags: /true/i.test(inputs['sync-semver-tags']),
-    author,
-  })
-
-  if (prBody.length > MAX_PR_BODY_SIZE_LIMITATION) {
-    const omissionText =
-      '. *Note: Part of the release notes have been omitted from this message, as the content exceeds the size limit*'
-    return _truncate(prBody, {
-      length: PR_BODY_TRUNCATE_SIZE,
-      omission: omissionText,
-    })
-  }
-
-  return prBody
-}
 
 const addArtifact = async (inputs, releaseId) => {
   const artifactPath = inputs['artifact-path']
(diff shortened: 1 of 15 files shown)


src/utils/releaseNotes.js
```diff
@@ -30,4 +36,44 @@ function getPrNumbersFromReleaseNotes(releaseNotes) {
...
+  })
+
+  if (prBody.length > MAX_PR_BODY_SIZE_LIMITATION) {
+    const omissionText =
+      '. *Note: Part of the release notes have been omitted from this message, as the content exceeds the size limit*'
+    return _truncate(prBody, {
+      length: PR_BODY_TRUNCATE_SIZE,
+      omission: omissionText,
+    })
+  }
```
simoneb, line 74: this logic contains a big bug

simoneb (changes requested):
see inline comment

src/utils/releaseNotes.js
sameer-coder: I have moved the metadata to the top of the template so it doesn't get truncated

simoneb: can we please truncate the release notes so that the truncation logic doesn't depend on where the release metadata is positioned in the body of the PR?

simoneb (commented):
I still don't like this solution too much. Let's keep the release meta in the template, but truncate the release notes instead.

Sorry for nitpicking, I think we should come up with a functional and clean solution.

sameer-coder:
> I still don't like this solution too much. Let's keep the release meta in the template, but truncate the release notes instead.
> 
> Sorry for nitpicking, I think we should come up with a functional and clean solution.

Ok. If the length of the release notes is more than 60K then I will truncate it to 60K. The rest of the template should take about 1.5K characters so the final PR body length will be always within limit.

Alternatively, if we first check that the rendered PR body exceeds 65536 and on basis of that decide if we need to truncate the release notes then that would mean running the template twice which is not ideal so I will not use this approach.

sameer-coder:
> I still don't like this solution too much. Let's keep the release meta in the template, but truncate the release notes instead.
> 
> Sorry for nitpicking, I think we should come up with a functional and clean solution.

I have changed the code so that if the length of the release notes is more than 60K then it will truncate it to 60K. The rest of the template should take about 1.5K characters max so the final PR body length will be always within limit.

Alternatively, if we first check that the rendered PR body exceeds 65536 and on basis of that decide if we need to truncate the release notes then that would mean running the template twice which is not ideal so I will not use this approach.

src/openPr.js
```diff
@@ -134,6 +105,10 @@ module.exports = async function ({ context, inputs, packageVersion }) {
       },
       inputs
     )
+    if (response?.status !== 201) {
+      const errMessage = response?.message || 'PR creation failed'
```
simoneb, line 109: are we sure that this is where the message will appear in case of an error?

src/utils/releaseNotes.js
```diff
@@ -30,4 +35,45 @@ function getPrNumbersFromReleaseNotes(releaseNotes) {
   return [...new Set(allPrNumbers)]
 }
 
+exports.getPRBody = (
```
simoneb, line 38: for consistency, let's declare a plain function and then export it at the bottom of this module, as we're doing with the other function

src/utils/releaseNotes.js
```diff
@@ -30,4 +35,45 @@ function getPrNumbersFromReleaseNotes(releaseNotes) {
...
+    id: draftRelease.id,
+    version: newVersion,
+    npmTag: inputs['npm-tag'],
+    opticUrl: inputs['optic-url'],
+  }
+
+  const draftReleaseBody = draftRelease?.body || ''
+  if (draftReleaseBody.length > PR_BODY_TRUNCATE_SIZE) {
+    const omissionText =
+      '. *Note: Part of the release notes have been omitted from this message, as the content exceeds the size limit*'
```
simoneb, line 59:
```suggestion
      '> Some of these release notes have been truncated to respect Pull Request body size limits'
```

simoneb (commented):
a few more comments

src/openPr.js
sameer-coder: yes, I tested this by simulating an error

src/utils/releaseNotes.js
```diff
@@ -30,4 +35,46 @@ function getPrNumbersFromReleaseNotes(releaseNotes) {
...
+    artifact,
+    syncTags: /true/i.test(inputs['sync-semver-tags']),
+    author,
+  })
+
+  return prBody
+}
+
 exports.getPrNumbersFromReleaseNotes = getPrNumbersFromReleaseNotes
+exports.getPRBody = getPRBody
```
simoneb, line 80: nit: let's slightly simplify this to module.exports = { ... }

test/releaseNotes.test.js
```diff
@@ -36,3 +42,54 @@ tap.test('Should return the correct PR numbers', async () => {
...
+
+  const truncatedPrBody = getPRBody(_template(tpl), {
+    newVersion: '1.0.0',
+    draftRelease: { id: 1, body: longPrBody },
+    inputs: [],
+    author: 'test',
+    artifact: null,
+  })
+  tap.ok(truncatedPrBody.length < 65536)
+  tap.ok(
```
simoneb, line 90: this is not the right way to assert here (it was probably a mistake in existing tests too). you should use the first argument provided to the test function instead

simoneb (approved):

PR #200, opened 2022-12-07
feat: enhance release notes between base and new version

### LINKED ISSUE #195: Improve reliability of release notes
At the moment GitHub seems to be quite unreliable in terms of how release notes for a release are generated, in the sense that sometimes (quite often), the release notes generated automatically in the draft release include changes from earlier releases, rather than from the latest release.

The way we're doing it now is:

1. generate draft release (it will/may have wrong release notes because gh picks up a wrong release as the previous release)
2. grab the body of the release and include it in the PR opened by this action

Because GH API [doesn't support specifying a previous release when creating a new release](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release), there don't seem to be ways to accomplish what we need, but on the other hand the API supports an operation to just generate the release notes (without saving them anywhere) and this API does allow specifying a previous release. See [here](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release).

So, an approach we may want to investigate is:

1. get the latest release. see [here](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release)
2. generate release notes for the next release using [this endpoint](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release) and providing the `previous_tag_name` input
3. [create the new release](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) switching off automatic release notes generation and providing the release notes in the body of the request

davideroffo:
It closes #195 

### Main changes:

- implemented the flow described in issue #195 
- created two new methods: `fetchLatestRelease` and `generateReleaseNotes`
- added unit tests for the new `releases.js` file
- fixed and improved existing the unit tests

### Testing:

In order to test my implementation I created a private repository and installed on it the Optic Release Automation App.
Then, I configured the `optic-release-automation-action`, specifying to use this branch and tried creating a release.

```
jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: nearform/optic-release-automation-action@feat/195-improve-reliability-of-release-notes
        with:
          github-token: ${{ secrets.github_token }}
          commit-message: "chore: release {version}"
          build-command: |
            npm ci
            npm run manifest
            npm run build
          artifact-path: dist
```

Below you can find the results.

#### First release:

Action logs:

> ** Starting Opening Release PR **
> Fetching latest release
> No previous releases found
> Creating draft release from commit: e194dec15cba077619271054dd9bc6e70150350a
> Draft release created successfully
> New version v1.0.0
> Artifact attached!
> ** Finished! **

PR description:

image

#### 2nd release (and next releases):

Action logs: 

> ** Starting Opening Release PR **
> Fetching latest release
> Latest release fetched successfully with tag: v2.0.5
> Generating release notes: [v2.0.5 -> v2.0.6]
> Release notes generated: v2.0.6
> Creating draft release from commit: f8877f2608452d730649d2853546a05109efb0f5
> Draft release created successfully
> New version v2.0.6
> Artifact attached!
> ** Finished! **

PR description:

image

### DIFF
--- a/src/openPr.js
+++ b/src/openPr.js
@@ -9,7 +9,7 @@ const { PR_TITLE_PREFIX } = require('./const')
 const { runSpawn } = require('./utils/runSpawn')
 const { callApi } = require('./utils/callApi')
 const transformCommitMessage = require('./utils/commitMessage')
-const { logInfo } = require('./log')
+const { logInfo, logWarning } = require('./log')
 const { attach } = require('./utils/artifact')
 const { getPRBody } = require('./utils/releaseNotes')
 const { fetchLatestRelease, generateReleaseNotes } = require('./utils/releases')
@@ -25,18 +25,23 @@ const addArtifact = async (inputs, releaseId) => {
   return artifact
 }
 
-const getReleaseNotes = async (inputs, newVersion) => {
-  const latestRelease = await fetchLatestRelease(inputs)
-  if (!latestRelease) {
-    return null
+const tryGetReleaseNotes = async (token, newVersion) => {
+  try {
+    const latestRelease = await fetchLatestRelease(token)
+    if (!latestRelease) {
+      return false
+    }
+    const { tag_name: baseVersion } = latestRelease
+    const releaseNotes = await generateReleaseNotes(
+      token,
+      newVersion,
+      baseVersion
+    )
+    return releaseNotes?.body
+  } catch (err) {
+    logWarning(err.message)
+    return false
   }
-  const { tag_name: latestVersion } = latestRelease
-  const releaseNotes = await generateReleaseNotes(
-    inputs,
-    newVersion,
-    latestVersion
-  )
-  return releaseNotes?.body || null
 }
 
 const createDraftRelease = async (inputs, newVersion, releaseNotes) => {
@@ -91,7 +96,9 @@ module.exports = async function ({ context, inputs, packageVersion }) {
 
   await run('git', ['push', 'origin', branchName])
 
-  const releaseNotes = await getReleaseNotes(inputs, newVersion)
+  const token = inputs['github-token']
+
+  const releaseNotes = await tryGetReleaseNotes(token, newVersion)
 
   const draftRelease = await createDraftRelease(
     inputs,
--- a/src/openPr.js
+++ b/src/openPr.js
@@ -27,7 +27,10 @@ const addArtifact = async (inputs, releaseId) => {
 
 const getReleaseNotes = async (inputs, newVersion) => {
   const latestRelease = await fetchLatestRelease(inputs)
-  const latestVersion = latestRelease ? latestRelease.tag_name : null
+  if (!latestRelease) {
+    return null
+  }
+  const { tag_name: latestVersion } = latestRelease
   const releaseNotes = await generateReleaseNotes(
     inputs,
     newVersion,
--- a/src/openPr.js
+++ b/src/openPr.js
@@ -82,7 +82,6 @@ module.exports = async function ({ context, inputs, packageVersion }) {
 
   const messageTemplate = inputs['commit-message']
   await run('git', ['checkout', '-b', branchName])
-  await run('git', ['pull', 'origin', branchName])
   await run('git', ['add', '-A'])
   await run('git', [
     'commit',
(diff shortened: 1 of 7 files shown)


sameer-coder (approved):

src/utils/releases.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    logError(err.message)
+    throw new Error(
+      `An error occurred while fetching the latest release: ${err.message}`
+    )
+  }
+}
+
+async function generateReleaseNotes(inputs, newVersion, latestVersion) {
```
simoneb, line 37: - no point in providing the whole inputs, this function only needs the token
- nit: let's not call it latestVersion, let's call it baseVersion

src/openPr.js
```diff
@@ -24,7 +25,21 @@ const addArtifact = async (inputs, releaseId) => {
   return artifact
 }
 
-const createDraftRelease = async (inputs, newVersion) => {
+const getReleaseNotes = async (inputs, newVersion) => {
+  const latestRelease = await fetchLatestRelease(inputs)
+  if (!latestRelease) {
+    return null
```
simoneb, line 31: let's not return nulls explicitly, it's not idiomatic. 

also, let's try to stick to a convention with functions that either return what they are meant to, or falsy. prefix their names with "try", e.g. tryGetReleaseNotes. this is to signify that either it does what it should, or it returns something (usually a falsy value) without throwing. the same name without the "try" prefix would instead throw an error if it can't do what we're asking it to do

src/openPr.js
```diff
@@ -24,7 +25,21 @@ const addArtifact = async (inputs, releaseId) => {
...
+  }
+  const { tag_name: latestVersion } = latestRelease
+  const releaseNotes = await generateReleaseNotes(
+    inputs,
+    newVersion,
+    latestVersion
+  )
+  return releaseNotes?.body || null
```
simoneb, line 39: no need for || null

src/openPr.js
```diff
@@ -24,7 +25,21 @@ const addArtifact = async (inputs, releaseId) => {
   return artifact
 }
 
-const createDraftRelease = async (inputs, newVersion) => {
+const getReleaseNotes = async (inputs, newVersion) => {
```
simoneb, line 28: as a consequence of generateReleaseNotes not needing the whole inputs, I believe this function also won't need the whole inputs object

src/utils/releases.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+const github = require('@actions/github')
+const { logInfo, logError } = require('../log')
+
+async function fetchLatestRelease(inputs) {
```
simoneb, line 6: this function also doesn't need the whole inputs object

src/utils/releases.js
```diff
@@ -0,0 +1,68 @@
...
+
+    logInfo(
+      `Latest release fetched successfully with tag: ${latestRelease.tag_name}`
+    )
+
+    return latestRelease
+  } catch (err) {
+    if (err.message === 'Not Found') {
+      logInfo(`No previous releases found`)
+      return null
```
simoneb, line 27: avoid returning nulls, just do a plain return which will return undefined instead

simoneb (changes requested):
Nothing major, but a few things to change/improve

src/openPr.js
```diff
@@ -24,7 +25,26 @@ const addArtifact = async (inputs, releaseId) => {
   return artifact
 }
 
-const createDraftRelease = async (inputs, newVersion) => {
+const tryGetReleaseNotes = async (token, newVersion) => {
+  try {
+    const latestRelease = await fetchLatestRelease(token)
+    if (!latestRelease) {
+      return false
```
simoneb, line 32:
```suggestion
      return
```

src/openPr.js
```diff
@@ -24,7 +25,26 @@ const addArtifact = async (inputs, releaseId) => {
...
+    const { tag_name: baseVersion } = latestRelease
+    const releaseNotes = await generateReleaseNotes(
+      token,
+      newVersion,
+      baseVersion
+    )
+    return releaseNotes?.body
+  } catch (err) {
+    logWarning(err.message)
+    return false
```
simoneb, line 43: no need to return anything

```suggestion
```

simoneb (approved):
a couple of comments but this LGTM. before merging, let's try it out in a couple of repos first.

PR #202, opened 2022-12-12
Auto version bump

sameer-coder:
WIP

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ on:
         default: 'patch'
         type: choice
         options:
+          - auto
           - patch
           - minor
           - major
--- a/action.yml
+++ b/action.yml
@@ -84,7 +84,7 @@ runs:
 
     - name: Version bump
       id: version-bump
-      if: ${{ github.event_name == 'workflow_dispatch' }}
+      if: ${{ github.event_name == 'workflow_dispatch' && inputs.semver !== 'auto' }}
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
         echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
--- a/action.yml
+++ b/action.yml
@@ -85,8 +85,8 @@ runs:
     - name: Version bump
       id: version-bump
       if: |
-        github.event_name == 'workflow_dispatch' ||
-        (github.event_name == 'workflow_dispatch' && inputs.semver != 'auto')
+        github.event_name == 'workflow_dispatch' &&
+        inputs.semver != 'auto'
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
         echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
--- a/action.yml
+++ b/action.yml
@@ -86,7 +86,7 @@ runs:
       id: version-bump
       if: |
         github.event_name == 'workflow_dispatch' ||
-        (github.event_name == 'workflow_dispatch' && inputs.semver !== 'auto')
+        (github.event_name == 'workflow_dispatch' && inputs.semver != 'auto')
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
         echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
--- a/action.yml
+++ b/action.yml
@@ -84,7 +84,9 @@ runs:
 
     - name: Version bump
       id: version-bump
-      if: ${{ github.event_name == 'workflow_dispatch' && inputs.semver !== 'auto' }}
+      if: |
+        github.event_name == 'workflow_dispatch' ||
+        (github.event_name == 'workflow_dispatch' && inputs.semver !== 'auto')
       run: |
         npm version --no-git-tag-version ${{ inputs.semver }}
         echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ const { logError } = require('./log')
 
 module.exports = async function ({ github, context, inputs, packageVersion }) {
   if (context.eventName === 'workflow_dispatch') {
-    return openPr({ context, inputs, packageVersion })
+    return openPr({ github, context, inputs, packageVersion })
   }
 
   if (context.eventName === 'pull_request') {
--- a/src/utils/bump.js
+++ b/src/utils/bump.js
@@ -1,0 +1,105 @@
+'use strict'
+const github = require('@actions/github')
+const { logDebug, logInfo } = require('../log')
+
+async function getBumpedVersion({ versionPrefix, token }) {
+  const { owner, repo } = github.context.repo
+  const data = await github.graphql(
+    `
+    query {
+      repository(owner: $repoOwner, name: $repoName) {
+        latestRelease{
+          tagName
+          tagCommit {
+            oid
+          }
+        }
+      }
+    }
+    `,
+    {
+      owner,
+      repo,
+    }
+  )
+
+  logInfo(`response from get latest release query ${JSON.stringify(data)}`)
+
+  const latestReleaseCommitSha = data?.repository?.latestRelease?.tagCommit?.oid
+  const latestReleaseTagName = data?.repository?.latestRelease?.tagName
+
+  if (!latestReleaseCommitSha || !latestReleaseTagName) {
+    logDebug(`response from get latest release query ${JSON.stringify(data)}`)
+    throw new Error(`Couldn't find latest release`)
+  }
+
+  const octokit = github.getOctokit(token)
+
+  const allCommits = await octokit.rest.repos.listCommits({
+    owner,
+    repo,
+    sha: latestReleaseCommitSha,
+    per_page: 100,
+    page: 1,
+  })
+  logInfo(`=-LOG-= ---> allCommits`, allCommits)
+
+  const isTagVersionPrefixed = latestReleaseTagName.includes(versionPrefix)
+
+  const currentVersion = isTagVersionPrefixed
+    ? latestReleaseTagName.replace(versionPrefix, '')
+    : latestReleaseTagName.replace(versionPrefix, 'v.') // default prefix
+
+  logInfo(`=-LOG-= ---> currentVersion`, currentVersion)
+
+  if (!currentVersion) {
+    logDebug(`response from get latest release query ${JSON.stringify(data)}`)
+    throw new Error(`Couldn't find latest version`)
+  }
+
+  return getVerionFromCommits(currentVersion, allCommits)
+}
+
+function getVerionFromCommits(currentVersion, commits = []) {
+  // Define a regular expression to match Conventional Commits messages
+  const commitRegex = /^(feat|fix|BREAKING CHANGE)(\(.+\))?:\s(.+)$/
+
+  // Define a mapping of commit types to version bump types
+  var versionBumpMap = {
+    'BREAKING CHANGE': 'major',
+    feat: 'minor',
+    fix: 'patch',
+  }
+
+  let [major, minor, patch] = currentVersion.split('.')
+  let isBreaking = false
+  let isMinor = false
+
+  for (const commit of commits) {
+    const match = commitRegex.exec(commit)
+    if (!match) continue
+
+    const type = match[1]
+
+    // Determine the version bump type based on the commit type
+    const bumpType = versionBumpMap[type]
+    if (!bumpType) continue
+
+    if (bumpType === 'major') {
+      isBreaking = true
+      break
+    } else if (bumpType === 'minor') {
+      isMinor = true
+    }
+  }
+
+  if (isBreaking) {
+    return `${major++}.0.0`
+  } else if (isMinor) {
+    return `${major}.${minor++}.0`
+  } else {
+    return `${major}.${minor}.${patch++}`
+  }
+}
+
+exports.getBumpedVersion = getBumpedVersion
--- a/src/utils/bump.js
+++ b/src/utils/bump.js
@@ -1,13 +1,13 @@
 'use strict'
-const github = require('@actions/github')
-const { logDebug, logInfo } = require('../log')
+const { logDebug } = require('../log')
+const { getOctokit } = require('@actions/github')
 
-async function getBumpedVersion({ versionPrefix, token }) {
-  const { owner, repo } = github.context.repo
+async function getBumpedVersion({ github, context, versionPrefix, token }) {
+  const { owner, repo } = context.repo
   const data = await github.graphql(
     `
-    query {
-      repository(owner: $repoOwner, name: $repoName) {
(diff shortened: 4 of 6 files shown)


src/utils/bump.js
```diff
@@ -0,0 +1,163 @@
...
+    owner,
+    repo,
+    commitDate: latestReleaseCommitDate,
+  })
+
+  if (!allCommits.length) {
+    throw new Error(`No commits found since last release`)
+  }
+
+  const bumpedVersion = getVerionFromCommits(latestReleaseTagName, allCommits)
```
maksimovicdanijel, line 33: **nitpick:** Typo in the function name. Should be `getVersionFromCommits`.

src/utils/bump.js
```diff
@@ -0,0 +1,163 @@
...
+    data?.repository?.latestRelease?.tagCommit?.committedDate
+
+  return {
+    latestReleaseCommitSha,
+    latestReleaseTagName,
+    latestReleaseCommitDate,
+  }
+}
+
+async function getCommitsSinceLatestRelease({
```
maksimovicdanijel, line 124: **opinion:** Maybe rename this function to `getCommitMessagesSinceLatestRelease` as that's what we are returning, an array of commit message.

test/auto-bump.test.js
```diff
@@ -0,0 +1,241 @@
...
+    context: DEFAULT_CONTEXT,
+  })
+
+  t.same(newversion, '3.0.0')
+})
+
+test('should bump minor major version if feat commit found', async t => {
+  const githubApiResponse = sinon.stub()
+
+  const commits = [
```
maksimovicdanijel, line 109: **opinion:** I would expand on the commits to include a fix as well so we can be sure that we are not bumping to minor version.

maksimovicdanijel (changes requested):
Typo in the function name, otherwise looks good.

README.md
```diff
@@ -74,6 +75,10 @@ The above workflow (when manually triggered) will:
 
 ![image](https://user-images.githubusercontent.com/2510597/140506212-4938e44d-0662-4dc5-9fb1-c3f59fe075a6.png)
 
+
+*Note*: If you choose the "auto" option for semver version updates, the action will attempt to determine the new version number based on the commit messages. For this option to work, the repository must use the conventional commits standard.
```
simoneb, line 79: this looks good, but maybe we should dedicate a specific section to the auto option rather than a simple note in the readme

action.yml
```diff
@@ -84,7 +84,9 @@ runs:
 
     - name: Version bump
       id: version-bump
-      if: ${{ github.event_name == 'workflow_dispatch' }}
+      if: |
+        github.event_name == 'workflow_dispatch' &&
+        inputs.semver != 'auto'
```
simoneb, line 89: I stopped reviewing a this point. Unless you think there's a good reason to do it the way you did, I don't think we should be changing the approach so much. I mean this step is bumping the package versions and the logic of the action assumes that it's bumped here and that there's a PACKAGE_VERSION output variable populated. If we move this logic elsewhere in case of an auto bump, we may inadvertently break invariants and assumptions that we're making elsewhere.

Now, I do understand that the auto logic is different, but why don't we execute it here instead of elsewhere? Basically, turn this step into: if auto { do something } else { do something else }, then the rest of the action remains unchanged

simoneb (changes requested):
See the main inline comment for the reason why I'm requesting changes

action.yml
sameer-coder: I agree, that would be cleaner approach. I will push an update soon

action.yml
```diff
@@ -83,12 +83,19 @@ runs:
...
+      with:
+        github-token: ${{ inputs.github-token }}
+        script: |
+          let newVersion = '${{ inputs.semver }}'
+          if ('${{ inputs.semver }}' == 'auto') {
+            const script = require('${{ github.action_path }}/dist/index.js')
+            newVersion = await script.getBumpedVersion({ github, context })
+          }
+          await exec.exec(`npm version --no-git-tag-version ${newVersion}`)
+          return require('./package.json').version
```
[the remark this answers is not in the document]
sameer-coder: Ok I will move this to a separate new function and export it to use here

src/utils/bump.js
```diff
@@ -0,0 +1,163 @@
...
+  parsedCommitDate.setSeconds(parsedCommitDate.getSeconds() + 1)
+
+  const data = await github.graphql(
+    `
+      query getCommitsSinceLastRelease($owner: String!, $repo: String!, $since: GitTimestamp!) {
+        repository(owner: $owner, name: $repo) {
+          defaultBranchRef {
+            target {
+              ... on Commit {
+                history(first: 100, since: $since) {
```
simoneb, line 140: what if there are more than 100?

src/utils/bump.js
```diff
@@ -0,0 +1,163 @@
...
+      since: parsedCommitDate,
+    }
+  )
+
+  const commitsList =
+    data?.repository?.defaultBranchRef?.target?.history?.nodes || []
+  return commitsList.map(c => c.message)
+}
+
+exports.getAutoBumpedVersion = getAutoBumpedVersion
```
simoneb, line 163: nit: module.exports = { getAutoBumpedVersion }

src/utils/bump.js
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!semver.valid(bumpedVersion)) {
+    throw new Error(`Invalid bumped version ${bumpedVersion}`)
+  }
+  return bumpedVersion
+}
+
+function getVersionFromCommits(currentVersion, commits = []) {
+  // Define a regular expression to match Conventional Commits messages
+  const commitRegex = /^(feat|fix|BREAKING CHANGE)(\(.+\))?:(.+)$/
```
simoneb, line 43: let's use an official conventionalcommits package to parse this and implement all this logic please!

simoneb (changes requested):

src/utils/bump.js
sameer-coder: Ok I was thinking about it as well but then decided to keep it simple. They have a conventionalcommits parser lib so I will just use that

src/index.js
```diff
@@ -15,3 +17,18 @@ module.exports = async function ({ github, context, inputs, packageVersion }) {
...
+      : inputs.semver
+
+  const run = runSpawn()
+  await run('npm', ['version', '--no-git-tag-version', newVersion])
+  return await run('npm', ['pkg', 'get', 'version'])
+}
+
+module.exports.runAction = runAction
+module.exports.getBumpedVersionNumber = getBumpedVersionNumber
+module.exports.getAutoBumpedVersion = getAutoBumpedVersion
```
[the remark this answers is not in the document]
sameer-coder: It was being used earlier in action.yaml but now that we moved all that code in index.js I will remove it.

simoneb (commented):
This repo https://github.com/conventional-changelog/conventional-changelog contains a tool called conventional-recommended-bump. Now I'm not saying that we should necessarily rely on that, but I believe the code here is doing much of what that tool does already

sameer-coder:
> This repo https://github.com/conventional-changelog/conventional-changelog contains a tool called conventional-recommended-bump. Now I'm not saying that we should necessarily rely on that, but I believe the code here is doing much of what that tool does already

I did try `conventional-recommended-bump` before using `@conventional-commits/parser` but couldn't get it to work(hence the delay in creating the PR). When I ran it, it wouldn't pickup all the commits after the previous release correctly. There is no way to explicitly pass list of commits as a parameter as well. I didn't want to spend too much time figuring out why that's happening but I can give it another shot if you think it's worth it.

src/utils/bump.js
```diff
@@ -0,0 +1,186 @@
...
+    return `${major}.${++minor}.0`
+  } else {
+    return `${major}.${minor}.${++patch}`
+  }
+}
+
+async function getLatestRelease({ github, owner, repo }) {
+  const data = await github.graphql(
+    `
+    query getLatestTagCommit($owner: String!, $repo: String!) {
```
simoneb, line 115: the name of this query doesn't look quite right

src/utils/bump.js
```diff
@@ -0,0 +1,186 @@
...
+  const latestReleaseCommitSha = data?.repository?.latestRelease?.tagCommit?.oid
+  const latestReleaseTagName = data?.repository?.latestRelease?.tagName
+  const latestReleaseCommitDate =
+    data?.repository?.latestRelease?.tagCommit?.committedDate
+
+  return {
+    latestReleaseCommitSha,
+    latestReleaseTagName,
+    latestReleaseCommitDate,
+  }
```
simoneb, line 142: let's simplify this object. we know it's the latestRelease, so we don't have to prepend latestRelease to all props!

src/utils/bump.js
```diff
@@ -0,0 +1,186 @@
...
+      }
+    }
+    `,
+    {
+      owner,
+      repo,
+    }
+  )
+
+  const latestReleaseCommitSha = data?.repository?.latestRelease?.tagCommit?.oid
```
simoneb, line 133: let's improve the error handling logic. throw here if the latestRelease could not be found and handle the error in the calling code so we don't have to check all 3 props there

simoneb (commented):
Overall LGTM, although I would be more comfortable reusing the logic inside the official libraries because there's a good chance we're missing something in our logic here. Nevertheless, let's address these comments for now. I still believe there must be a way to use that library, I'm confident that it must be relied upon by others, hence it should work.

src/utils/bump.js
```diff
@@ -0,0 +1,186 @@
...
+    !latestReleaseCommitDate
+  ) {
+    throw new Error(`Couldn't find latest release`)
+  }
+
+  const allCommits = await getCommitMessagesSinceLatestRelease({
+    github,
+    owner,
+    repo,
+    commitDate: latestReleaseCommitDate,
```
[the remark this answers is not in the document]
sameer-coder: If we use the commit sha then using the graphql API there is no easy way to get the list of commits after a particular sha. I will explore a bit more to see how we can avoid using `since` and do it using the sha.

> commits with an older date could land in the repository after commits with a newer date
I _guess_ using `since` should take care of this automatically but I will try this out to see how the code behaves in this scenario

sameer-coder:
Using the `conventional-recommended-bump` npm package now to do the version bump

action.yml
```diff
@@ -83,12 +83,15 @@ runs:
...
-      run: |
-        npm version --no-git-tag-version ${{ inputs.semver }}
-        echo "PACKAGE_VERSION=$(echo $(node -p "require('./package.json').version"))" >> $GITHUB_OUTPUT
-      shell: 'bash'
+      if: github.event_name == 'workflow_dispatch'
+      with:
+        result-encoding: string
+        github-token: ${{ inputs.github-token }}
+        script: |
+          const { getBumpedVersionNumber } = require('${{ github.action_path }}/dist/index.js')
```
simoneb, line 93: let's call this function `bumpVersion` which it more telling that it doesn't just get the version number, it also bumps it!

src/index.js
```diff
@@ -2,9 +2,19 @@
 
 const openPr = require('./openPr')
 const release = require('./release')
+const { runSpawn } = require('./utils/runSpawn')
 const { logError } = require('./log')
+const core = require('@actions/core')
+const util = require('util')
+const config = require('conventional-changelog-conventionalcommits')
```
simoneb, line 9: this is a slightly too generic name

src/index.js
```diff
@@ -15,3 +25,27 @@ module.exports = async function ({ github, context, inputs, packageVersion }) {
...
+    inputs.semver === autoInput ? await getAutoBumpedVersion() : inputs.semver
+
+  const run = runSpawn()
+  await run('npm', ['version', '--no-git-tag-version', newVersion])
+  return await run('npm', ['pkg', 'get', 'version'])
+}
+
+async function getAutoBumpedVersion() {
+  try {
+    const { releaseType } = await conventionalRecommendedBumpAsync({ config })
```
simoneb, line 40: why don't we inline the require so we don't even require the package unless this function is executed?

simoneb: ok, this is nice because it's simple, but do we know what it's actually doing? I mean how do they know which commits to pick?

simoneb: @sameer-coder ^ this is the comment I was referring to. we need a way to specify the base tag/commit from which to generate the list of commits to consider. the package I believe currently doesn't support this, let's send them a PR

simoneb:
@davideroffo can you please review this?

davideroffo (commented):
LGTM, I left a few minor comments.

package.json
```diff
@@ -19,6 +19,7 @@
     "@actions/core": "^1.10.0",
     "@actions/github": "^5.1.1",
     "adm-zip": "^0.5.10",
+    "conventional-changelog-monorepo": "github:nearform/conventional-changelog",
```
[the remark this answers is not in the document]
sameer-coder: There is already a ticket created as a reminder in the nearform/conventional-changelog repo

action.yml
```diff
@@ -72,6 +72,9 @@ inputs:
   prerelease-prefix:
     description: 'A prefix to apply to the prerelease version number'
     required: false
+  base-tag:
```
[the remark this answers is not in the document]
sameer-coder: yes base tag acts like the `from` tag for auto-version-bump. I think `base-tag` makes more sense for auto bump because internally we use the latest tag if `base-tag` input is not provided and conventional-changelog package also uses tags for the bump. what do you think?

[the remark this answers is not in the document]
davideroffo: Yes, let's use it. I'll use the same one in my task as well.

package.json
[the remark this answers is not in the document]
sameer-coder: Sure. Created a new ticket [here](https://github.com/nearform/optic-release-automation-action/issues/231)

davideroffo (approved):

davideroffo:
@sameer-coder I noticed that the `base-tag` parameter has not been included in the `Inputs` section of README.md.

davideroffo (approved):

src/index.js
```diff
@@ -15,3 +25,55 @@ module.exports = async function ({ github, context, inputs, packageVersion }) {
...
+
+    const { releaseType = 'patch' } = await conventionalRecommendedBumpAsync({
+      baseTag: tag,
+      config: conventionalCommitsConfig,
+    })
+    logInfo(`Auto generated release type is ${JSON.stringify(releaseType)}`)
+    return releaseType
+  } catch (error) {
+    core.setFailed(error.message)
+    throw error
```
simoneb, line 72: this error handling seems unnecessary. let's make sure error handling is done properly in the root of the action, which means that calling setFailed here and rethrowing the error shouldn't be needed

simoneb (changes requested):
a few comments

.github/workflows/release.yml
```diff
@@ -32,3 +33,4 @@ jobs:
           base-tag: ${{ github.event.inputs.baseTag }}
           sync-semver-tags: 'true'
           build-command: npm ci
+          base-tag: ${{ github.event.inputs.baseTag }}
```
davideroffo, line 36: this one is a duplicate

src/index.js
sameer-coder: This is one is called directly from `action.yaml`. I too think we can remove this error handling since if this function throws the action should fail and report error message automatically.

simoneb (approved):
LGTM

PR #205, opened 2022-12-14
refactor runSpawn to use @actions/exec

### LINKED ISSUE #190: Replace runSpawn calls with @actions/exec
We should replace runSpawn calls with [@actions/exec](https://github.com/actions/toolkit/tree/main/packages/exec) since it's the official way to run bash commands in actions.

maksimovicdanijel:
Rewrites the` runSpawn` utility function to use `@actions/exec` package, the recommended way of executing scripts in actions. It renames the function to `execWithOutput` and changes the function signature to be more appropriate for usage.

It also adds missing tests so we can have 100% test coverage 🎉 

Closes #190


simoneb:
Even if approved, please wait for my review on this one

simoneb (commented):
I can see that this function is exported from the package but I don't see it documented. How comfortable are we that we're not building upon internal APIs that may stop working any time?

sameer-coder:
Looks good other than the `getExecOutput` usage. I would go with the standard documented functions.
Also, we can add a test to improve the coverage.

maksimovicdanijel:
> How comfortable are we that we're not building upon internal APIs that may stop working any time?
Comfortable. I've started rewriting `runSpawn` function using the plain `exec` function with the listeners option and I saw the `getExecOutput` function which looked quite similar to what I wrote. `getExecOutput` function is using `exec` with predefined options, it's just a thin wrapper around it.

simoneb:
> > How comfortable are we that we're not building upon internal APIs that may stop working any time?
> 
> Comfortable. I've started rewriting `runSpawn` function using the plain `exec` function with the listeners option and I saw the `getExecOutput` function which looked quite similar to what I wrote. `getExecOutput` function is using `exec` with predefined options, it's just a thin wrapper around it.

Ok but this is not what I'm saying. What gives you confidence that this function will not disappear or change in behavior and thereby breaking our action without us even noticing?

simoneb (changes requested):
I would do two changes here:

- because the implementation of this function has changed, we don't need to preserve the existing signature, since it doesn't make much sense for the new function. I would both rename the module and function and change the signature so it doesn't export a function which returns a function, which is now unnecessary. In other words, let's take the opportunity to change the runSpawn module in a way that doesn't carry any leftovers from the previous implementation
- There's absolutely no testing for this. If you replace the implementation of this function with `return ''` all tests will succeed. Let's please cover this module with tests and, since we're there and unless it's too much effort, enforce 100% code coverage in tap (we're at around 91% now). If it's too much effort to reach 100% in this PR compared to the original purpose of this PR, let's do it in another PR

maksimovicdanijel:
> because the implementation of this function has changed, we don't need to preserve the existing signature, since it doesn't make much sense for the new function

@simoneb I was also thinking about this but didn't want to make a disruptive change in the repo as I'm quite new to it. I'll refactor the module itself and add proper testing. 

I've also opened an issue on the `toolkit` repo asking about the `getExecOutput` function https://github.com/actions/toolkit/issues/1267

simoneb:
@maksimovicdanijel perfect approach, thanks

.taprc
```diff
@@ -1,4 +1,4 @@
-statements: 90
-lines: 90
-functions: 78
-branches: 90
+statements: 100
```
simoneb, line 1: 100 should be the default. try removing this file if there's no other option we're using

simoneb (approved):
LGTM, nice job

.taprc
maksimovicdanijel: I've removed the file, please re-approve. Thanks!

PR #208, opened 2022-12-16
Revert "refactor runSpawn to use @actions/exec (#205)"

maksimovicdanijel:
This reverts commit 66727ac3c8a62ed681342c10e6b04a47a85e6018.


simoneb (approved):
please double check that this really only reverts your changes and nothing else before merging

maksimovicdanijel:
> please double check that this really only reverts your changes and nothing else before merging

Yes, these were the changes I made. It looks good 👍

PR #223, opened 2022-12-22
Change max PR body size

### LINKED ISSUE #224: Server error when creating PR
The GitHub API is throwing an error with the message `Server error` and status code `502` when creating PR.

https://github.com/nearform/slack-channel-github-team-sync/actions/runs/3758018143/jobs/6385862973

sameer-coder:
It seems the 60000 characters limit we had for release notes is not working as expected. The GitHub API is throwing an error with the message `Server error` and status code`502` if the limit is more than 30000.

Note: I found this limit by trial and error and did not find any documentation regarding this problem.

Closes #224

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -27619,7 +27619,7 @@ const _truncate = __nccwpck_require__(4436)
 
 const md = __nccwpck_require__(8561)()
 
-const PR_BODY_TRUNCATE_SIZE = 60000
+const PR_BODY_TRUNCATE_SIZE = 30000
 
 function getPrNumbersFromReleaseNotes(releaseNotes) {
   const parsedReleaseNotes = md.parse(releaseNotes, {})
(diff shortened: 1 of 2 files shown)


davideroffo:
LGTM but I'm wondering if it was working in the past with the pre-existing 60k limit.

Could you please try the action with the release 4.1.1? I'm just curious to understand if this behaviour has been caused by the "Generate Release Notes" feature I implemented, and the 4.1.1 is not including it.

If this is still happening on that release it means that probably that limit has been changed (from GitHub side?). Maybe we should investigate a bit on this side.

sameer-coder:
yeah, I too am not fully convinced that this is a good solution. This is not related to the "Generate Release Notes" feature. Even if you try to create a PR using curl with a large body it will give the same error. I suspect something from the GitHub side has changed and that seems to have reduced the max limit.

simoneb:
Folks let's try to get to the bottom of this please. We got it wrong once, let's stop guessing and come up with a robust solution instead.

sameer-coder:
Created this [bug](https://github.com/nearform/optic-release-automation-action/issues/229) to revert this change based on the response from GitHub team

PR #236, opened 2023-01-09
Remove release notes related to optic release

### LINKED ISSUE #219: Remove from the changelog the commit related to the last release
If we check the changelog _(what's changed section)_ when a release is created, we notice that a commit related to the last release is included.
We don't want this kind of log in the changelog.

Example: https://github.com/nearform/optic-release-automation-app/releases/tag/v0.4.3

### Current behaviour

image

### Expected behaviour

Untitled

sameer-coder:
This PR removes the notes generated by the optic release automation from the changelog.

Closes #219

### DIFF
--- a/src/utils/releases.js
+++ b/src/utils/releases.js
@@ -63,7 +63,8 @@ async function generateReleaseNotes(token, newVersion, baseVersion) {
 
     logInfo(`Release notes generated: [${baseVersion} -> ${newVersion}]`)
 
-    return excludeUnwantedReleaseNotes(releaseNotes)
+    const filteredReleaseNotes = excludeUnwantedReleaseNotes(releaseNotes)
+    return filteredReleaseNotes
   } catch (err) {
     throw new Error(
       `An error occurred while generating the release notes: ${err.message}`
--- a/src/utils/releases.js
+++ b/src/utils/releases.js
@@ -1,7 +1,9 @@
 'use strict'
 
 const github = require('@actions/github')
-const { logInfo } = require('../log')
+const { logInfo, logWarning } = require('../log')
+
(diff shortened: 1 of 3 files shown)


src/utils/releases.js
```diff
@@ -48,7 +63,7 @@ async function generateReleaseNotes(token, newVersion, baseVersion) {
 
     logInfo(`Release notes generated: [${baseVersion} -> ${newVersion}]`)
 
-    return releaseNotes
+    return excludeUnwantedReleaseNotes(releaseNotes)
```
davideroffo, line 66: maybe will be more readable if we assign the return value of `excludeUnwantedReleaseNotes()` to a variable and then return the variable.?

davideroffo (approved):
LGTM just a comment

sameer-coder:
> I don't think we should go in this direction. We should first figure out why this happens and why it doesn't happen other times. Removing entries manually more or a workaround than a solution.

I tried to find the actual cause of why it is happening. In my observation, when we call the GitHub API to get release notes, the entry related to the previous release it always there as the last entry. At times it comes back as the first entry or somewhere in the middle which is when it shows up in the changelog. When we send the release notes to GitHub to create a draft release that is when it decides whether to include it or not.

I know this sounds fuzzy but I haven't been able to pin point the exact reason why it is happening since most of it process happens on the GitHub side.

simoneb:
I understand this is confusing, but I would start from comparing places where this didn't happen to places where it happened, to see if we can pinpoint anything different that we may be doing in those repos.

simoneb:
Let's close this PR as per the outcome of the linked issue please

PR #237, opened 2023-01-10
feat: add base release selection support

### LINKED ISSUE #215: Support selecting what is the base release for a release
With the support for prereleases, we should now be able to provide as an input to the action the identifier of the base release.

- Say you have release 1.0.0 then you do a couple of prereleases of version 2, e.g. 2.0.0-beta.1 and 2.0.0-beta.2
- Those beta releases will have their own changelogs, and those changelogs you will probably want to include when you release the official 2.0.0
- Having an input called e.g. `base-release`, which in this case should be `1.0.0` will be used to generate release notes starting from the last official version, instead of the last beta version

davideroffo:
Closes #215 

### Main changes:

This PR adds the ability to select a base release (specifying the`tag`) from which to start. It is useful in order to generate a `changelog` containing all the changes between the base release and the newly created release.

- added a new `base-tag` input that will be populated in the `workflow_dispatch` inputs
- updated the `openPr` logic in order to check if the `base-tag` input has been populated and in this case tries to fetch the specified release (just to check if it exists through the new method `fetchReleaseByTag`). If not populated the standard behaviour will be executed (release notes will be generated starting from the latest release). If the `tag` is specified but it doesn't exist an error will be thrown
- added tests
- updated the `release.yml` workflow file in order to handle the `base-tag` input
- updated the `README.md` file


src/utils/releases.js
```diff
@@ -56,7 +56,35 @@ async function generateReleaseNotes(token, newVersion, baseVersion) {
   }
 }
 
+async function fetchReleaseByTag(token, tag) {
```
Ceres6, line 59: Nit: Maybe I'm missing something but it seems you're not actually using the return but rather checking if it doesn't throw. Maybe it could not return and be named something like `ensureReleaseTagExists` so the intention is more clear

Ceres6 (approved):

sameer-coder (approved):

action.yml
```diff
@@ -72,6 +72,9 @@ inputs:
   prerelease-prefix:
     description: 'A prefix to apply to the prerelease version number'
     required: false
+  base-tag:
+    description: 'Specifies the starting release base tag and the changelog will include all the changes made in releases that occur after the base release, up until the latest release.'
```
simoneb, line 76: this description is a bit confusing. what the base-tag should do is use that instead of the latest as a base, but everything else is unchanged (I'm referring to the "up until the latest release" part)

src/openPr.js
```diff
@@ -25,13 +29,18 @@ const addArtifact = async (inputs, releaseId) => {
...
-    if (!latestRelease) {
-      return
+    let baseVersion = baseRelease
+    if (!baseRelease) {
+      const latestRelease = await fetchLatestRelease(token)
+      if (!latestRelease) {
+        return
+      }
+      const { tag_name } = latestRelease
+      baseVersion = tag_name
```
simoneb, line 41: this logic is a bit confusing, let's try to write it in a more intuitive way or extract in a function

src/openPr.js
```diff
@@ -80,6 +89,12 @@ module.exports = async function ({ context, inputs, packageVersion }) {
     throw new Error('packageVersion is missing!')
   }
 
+  const token = inputs['github-token']
+
+  const baseReleaseTag = inputs['base-tag']
+  if (baseReleaseTag) {
+    await fetchReleaseByTag(token, baseReleaseTag)
+  }
```
simoneb, line 97: not sure I understand what this does. why would invoking a function and doing nothing with the result achieve anything?

simoneb: ah I see what this does and @Ceres6 comment below. ok no we need to do this differently. either we assert the existence, as @Ceres6 suggested, or we actually retrieve the release object and pass it around. the second option is probably more work but I have a feeling it would be a better solution

README.md
```diff
@@ -209,6 +211,7 @@ Please note that in case of a prerelease the `sync-semver-tags` input will be tr
...
 | `prerelease-prefix`       | No       | A prefix to apply to the prerelease version number.                                                                                                                                                                                                                                         |
+| `base-tag`       | No       | Specifies the starting release tag and the changelog will include all the changes made in releases that occur after the base release, up until the latest release. For example, if the base release is specified as `v1.0.0` and the latest release is `v2.0.0`, the release changelog will include all changes made in releases between `v1.0.0` and `v2.0.0`                                                                                                                                                                                                                                         |
 ## Motivation
```
simoneb, line 215: as per the previous comment in the action.yml input, this description needs improvement as it's quite confusing

simoneb (changes requested):

davideroffo:
Thanks, @Ceres6 and @simoneb for your review.

I addressed your change requests and I tested these scenarios (and everything is working fine):

- trigger the release with an invalid `base-tag` -> release fails and throw an error as expected ✅
- trigger the release without a `base-tag` -> release created used the latest release as base tag ✅
- trigger the release with a previous `base-tag` -> release created using the specified previous `base-tag` ✅
- trigger a first release (without prior releases available) -> the release is created successfully ✅

simoneb (commented):
Code looks much better, thanks, but the description of the new action input is still not clear. We must clarify that this is used for release notes alone, and has no functional implications otherwise. This may be obvious to you because you implemented it, but to a user this needs to be made clearer

simoneb (approved):

PR #247, opened 2023-01-18
Fix failing npm after git spawn change

matijagaspar:
Fixes: #248

The issue was that:
```javascript
const code = await exec(cmd, args, options)
```
Will throw an error if it fails to execute and the error will not contain the console output of the exec. I wrapped the bit into a try catch, so the underlying console output can be collected and put into the error message that is later handled by the outer function


sameer-coder (commented):
LGTM

sameer-coder (approved):

simoneb (approved):

Issue #285, opened 2023-04-19
Add support for provenance


simoneb:
https://github.blog/2023-04-19-introducing-npm-package-provenance/

Things to consider:

- does it work on older npm versions? probably not

AlanSl:
Regarding the "older versions of NPM" question: 

Running `npm publish --provenance` in 9.4 (before the release in 9.5) just ignores the flag, it doesn't cause it to fail. NPM 9.5.1 is already the default in `ubuntu-latest` so most CI jobs should be 

I'm assuming we want provenance to be an opt-in option defined something like this?

```yml
      - uses: nearform-actions/optic-release-automation-action
        with:
          semver: ${{ github.event.inputs.semver }}
          commit-message: 'chore: release {version}'
          provenance: true
```

So we could just ignore NPM version, but maybe we want to abort if a user opted in to provenance and doesn't realise their CI is using a too-old version of NPM, rather than letting the publish succeed and the provenance aspect fail silently? With something like this maybe:

```js
if (inputs['provenance']) {
  const npmVersion = getNpmVersion()
  if (!semver.satisfies(npmVersion, '>=9.5.0') {
    throw new Error(`Provenance option requires action to run on NPM >=9.5.0. You are using NPM ${npmVersion}`)
  }
}
```

AlanSl:
Regarding actually adding provenance support:

Unfortunately, simply adding `--provenance` flag to the exec call within `publishToNpm.js` fails with this error:

```
npm notice Publishing to https://registry.npmjs.org/ with tag latest and default access
npm ERR! code EUSAGE
npm ERR! Automatic provenance generation not supported outside of GitHub Actions
```

We're calling `npm publish` in node from GitHub Actions' own `@actions/exec`, from within a GitHub Action. You'd think this would be sufficiently within GitHub Actions, but apparently not.

I've had a look at the example NPM package that successfully published using provenance linked to from the blog, and they appeared to use `changesets`, which uses `publishScript = core.getInput("publish")` from `@actions/core`. Presumably that's more wired in to the GitHub Actions publish API than `exec` which could run anything?

So if there are no objections or other suggestions, I'll see if it's possible to refactor our publish step to use `core.getInput("publish")` instead of `exec`.

AlanSl:
[...]

PR #294, opened 2023-05-02
feat(issue 285): Add provenance support

AlanSl:
closes https://github.com/nearform-actions/optic-release-automation-action/issues/285

Adds [provenance](https://github.blog/2023-04-19-introducing-npm-package-provenance/) support, with:

 - Guards against silent failure or known misleading errors from NPM
 - Basic usage documentation
 - Unit tests and basic integration-like test cases in "release" tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ If `provenance: true` is added to your `release.yml`'s **inputs**, NPM will [gen
 NPM has some internal [requirements](https://docs.npmjs.com/generating-provenance-statements#prerequisites) for generating provenance. Unfortunately as of May 2023, not all are documented by NPM; some key requirements are:
 
 - `id-token: write` must be added to your `release.yml`'s **permissions**
-- NPM must on version 9.5.0 or greater (this will be met if our recommended `runs-on: ubuntu-latest` is used)
+- NPM must be on version 9.5.0 or greater (this will be met if our recommended `runs-on: ubuntu-latest` is used)
 - NPM has some undocumented internal requirements on `package.json` completeness. For example, the [repository field](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository) is required, and some NPM versions may require its `"url"` property to match the format `"git+https://github.com/user/repo"`.
 
 If any requirements are not met, the release will be aborted before publishing the new version, and an appropriate error will be shown in the actions report. The release commit can be reverted and the action re-tried after fixing the issue highlighted in the logged error.
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ NPM will [generate a provenance statement](https://docs.npmjs.com/generating-pro
    - Add `provenance: true` to your list of **inputs**
    - Add `id-token: write` to your list of **permissions**
  - Ensure your CI runner uses NPM >= 9.5.0 (should be the default if Node >= 18)
- - Ensure your `package.json` is complete and correct. NPM will cancel the release with specific errors if it finds a problem.Requirements may change in future NPM releases but include things like a `"repository"` field with `"url"` property matching the format `"git+https://github.com/user/repo"`.
+ - Ensure your `package.json` is complete and correct. NPM will cancel the release with specific errors if it finds a problem. Requirements may change in future NPM releases but include things like a `"repository"` field with `"url"` property matching the format `"git+https://github.com/user/repo"`.
 
 If `provenance: true` is in your YML inputs and any condition isn't met, the release will cancel with an error. 
 
@@ -254,7 +254,7 @@ If you run a release without `provenance: true` when a previous release had prov
 | `version-prefix`       | No       | A prefix to apply to the version number, which reflects in the tag and GitHub release names. <br /> (_Default: 'v'_)                                                                                                                                                                                                                                         |
 | `prerelease-prefix`       | No       | A prefix to apply to the prerelease version number.                                                                                                                                                                                                                                         |
 | `base-tag`       | No       | Choose a specific tag release for your release notes. This input allows you to specify a base release (for example, v1.0.0) and will include all changes made in releases between the base release and the latest release. This input is only used for generating release notes and has no functional implications on the rest of the workflow.                                                                                                                                                                                                                                         |
-| `provenance`| No    | Set as true to have NPM [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements. Also <br /> (_Default: `false`_)                                                                  |
+| `provenance`| No    | Set as true to have NPM [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements.<br /> (_Default: `false`_)                                                                  |
 
 
 ## Motivation
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ This action allows you to automate the release process of your npm modules, apps
   - [Example](#example-1)
 - [How to add a build step to your workflow](#how-to-add-a-build-step-to-your-workflow)
 - [Prerelease support](#prerelease-support)
+- [Provenance](#provenance)
 - [Inputs](#inputs)
 - [Motivation](#motivation)
 - [Playground / Testing](#playground--testing)
@@ -217,6 +218,20 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
 
 Please note that in case of a prerelease the `sync-semver-tags` input will be treated as `false`, even if it's set to `true`. This because we don't want to update the main version tags to the latest prerelease commit but only to the latest official release.
 
+## Provenance
+
+NPM will [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements) for your releases if the following conditions are met:
+
+ - In the `.yml` file that configures your release action:
+   - Add `provenance: true` to your list of **inputs**
+   - Add `id-token: write` to your list of **permissions**
+ - Ensure your CI runner uses NPM >= 9.5.0 (should be the default if Node >= 18)
+ - Ensure your `package.json` is complete and correct. will cancel the release with specific errors if it finds a problem. Requirements may change in future NPM releases but include things like a `"repository"` field with `"url"` property matching the format `"git+https://github.com/user/repo"`.
+
+If `provenance: true` is in your YML inputs and any condition isn't met, the release will cancel with an error. 
+
+If you run a release without `provenance: true` when a previous release had provenance, the provenance banner on your package's NPM landing page will be removed but the banner will remain on the subpage for the previous release.
+
 ## Inputs
 
 | Input          | Required | Description                                                                                                                                                                                |
@@ -239,6 +254,8 @@ Please note that in case of a prerelease the `sync-semver-tags` input will be tr
 | `version-prefix`       | No       | A prefix to apply to the version number, which reflects in the tag and GitHub release names. <br /> (_Default: 'v'_)                                                                                                                                                                                                                                         |
 | `prerelease-prefix`       | No       | A prefix to apply to the prerelease version number.                                                                                                                                                                                                                                         |
 | `base-tag`       | No       | Choose a specific tag release for your release notes. This input allows you to specify a base release (for example, v1.0.0) and will include all changes made in releases between the base release and the latest release. This input is only used for generating release notes and has no functional implications on the rest of the workflow.                                                                                                                                                                                                                                         |
+| `provenance`| No    | Set as true to have NPM [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#Provenance) for requirements. Also <br /> (_Default: `false`_)                                                                  |
+
 
 ## Motivation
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -217,6 +218,20 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
+
+ - In the `.yml` file that configures your release action:
+   - Add `provenance: true` to your list of **inputs**
+   - Add `id-token: write` to your list of **permissions**
+ - Ensure your CI runner uses NPM >= 9.5.0 (should be the default if Node >= 18)
+ - Ensure your `package.json` is complete and correct. NPM will cancel the release with specific errors if it finds a problem. Requirements may change in future NPM releases but include things like a `"repository"` field with `"url"` property matching the format `"git+https://github.com/user/repo"`.
+
+If `provenance: true` is in your YML inputs and any condition isn't met, the release will cancel with an error. 
+
+If you run a release without `provenance: true` when a previous release had provenance, the provenance banner on your package's NPM landing page will be removed but the banner will remain on the subpage for the previous release.
```
simoneb, line 233: this is certainly good to know, but is it really something that we need to document or shall we just redirect the user to the official docs? I mean we are not the owners of this information, and it may change

src/release.js
```diff
@@ -95,9 +99,23 @@ module.exports = async function ({ github, context, inputs }) {
   try {
     const opticToken = inputs['optic-token']
     const npmToken = inputs['npm-token']
+    const provenance = /true/i.test(inputs['provenance'])
```
simoneb, line 102: minor: for boolean inputs @actions/core contains a utility (but I believe what you're doing here works just fine and I'm fine with it

src/utils/provenance.js
```diff
@@ -0,0 +1,70 @@
...
+  // There are various other provenance requirements, such as specific package.json properties, but these
+  // may change in future NPM versions, and do fail with meaningful errors, so we let NPM handle those.
+}
+
+/**
+ * Gets npm version via `npm -v` on the command line.
+ * Split out as its own export so it can be easily mocked in tests.
+ */
+async function getNpmVersion() {
+  return await execWithOutput('npm', ['-v'])
```
simoneb, line 62: nit: not necessary to await the last async function invoked by an async function, it can simply be returned

src/utils/provenance.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
```
simoneb, line 1: generally speaking I'm not a fan of programming against specific versions of tools we use, npm in this case, but you did such a good job here at covering the various cases and doing the research that it would be a pity to undo it. so thanks for all this work, very nice

src/utils/execWithOutput.js
```diff
@@ -11,16 +11,14 @@ const { exec } = require('@actions/exec')
...
   let output = ''
   let errorOutput = ''
 
   const stdoutDecoder = new StringDecoder('utf8')
   const stderrDecoder = new StringDecoder('utf8')
 
-  const options = {
-    silent: false,
-  }
+  options.silent = false
```
simoneb, line 21: minor: can we implement this differently so if for whatever reason the caller specifies that silent should be true, we don't override it?

test/release.test.js
```diff
@@ -200,6 +216,73 @@ tap.test('Should publish to npm without optic', async () => {
   })
 })
 
+tap.only(
```
simoneb, line 219: remove .only, here and in the other tests

src/utils/provenance.js
```diff
@@ -0,0 +1,70 @@
...
+ * due to an NPM bug that existed between 9.5.0 and 9.6.1.
+ * As of April 2023, this would affect anyone whose CI is set to Node 18 (which defaults to NPM 9.5.1).
+ */
+function checkPermissions(npmVersion) {
+  // Bug was fixed in this NPM version - see https://github.com/npm/cli/pull/6226
+  const correctNpmErrorVersion = '>=9.6.1'
+
+  if (
+    // Same test condition as in fixed versions of NPM
+    !process.env.ACTIONS_ID_TOKEN_REQUEST_URL &&
```
simoneb, line 31: is this what tells us that we have the permissions? Q: do we really need to test against this? can't we just let it fail if the user did not configure the permission? e.g. what if GH changes the implementation and this variable is gone?

README.md
```diff
@@ -217,6 +218,20 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
 Please note that in case of a prerelease the `sync-semver-tags` input will be treated as `false`, even if it's set to `true`. This because we don't want to update the main version tags to the latest prerelease commit but only to the latest official release.
 
+## Provenance
+
+NPM will [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements) for your releases if the following conditions are met:
+
+ - In the `.yml` file that configures your release action:
+   - Add `provenance: true` to your list of **inputs**
+   - Add `id-token: write` to your list of **permissions**
+ - Ensure your CI runner uses NPM >= 9.5.0 (should be the default if Node >= 18)
```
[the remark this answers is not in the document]
AlanSl: The current approach appears to have been that we _do_ delegate this responsibility to the user, but we provide them with a working example config to follow (a `.github/workflows/release.yml` containing `runs-on: ubuntu-latest`). The readme said:

> It is important to be aware that you are responsible for:
> 
> 1. The context where the command is executed. <br /> The command will be executed using the current [GitHub runner](https://github.com/actions/runner#github-actions-runner) configuration (e.g. the one set on `runs-on`).
> You can customize it by...

But we can probably drop or amend this line since our recommended image [already provides NPM 9.5.1](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md), so they'll already get a valid version unless they deviated from our guidance. If they did for some reason, we give an informative error message, then beyond that, it's on them.

action.yml
```diff
@@ -75,6 +75,11 @@ inputs:
   base-tag:
     description: 'This input allows you to specify a base release and will include all changes made in releases between the base release and the latest release'
     required: false
+  provenance:
+    description: 'If true, NPM >9.5 will attempt to generate and display a "provenance" badge (https://github.blog/2023-04-19-introducing-npm-package-provenance/).'
```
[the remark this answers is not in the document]
AlanSl: 👍 done

src/utils/provenance.js
AlanSl: > generally speaking I'm not a fan of programming against specific versions of tools we use, npm in this case

Agreed, me too. In this case I think it's worth making an exception because our recommended config gives users a version of NPM containing a bug that we know actively misinforms them about their problem.

> can't we just let it fail if the user did not configure the permission?

Not when NPM version < 9.6.1 (i.e. everyone using `ubuntu-latest` as per our recommended config): that linked NPM bug causes them to get a very misleading error message _"Automatic provenance generation not supported outside of GitHub Actions"_ which will send them in entirely the wrong direction. 

> what if GH changes the implementation and this variable is gone?

This check is already limited to only apply on NPM < 9.6.1 in the next line. Every NPM version > 9.5.0 includes this same check on `!process.env.ACTIONS_ID_TOKEN_REQUEST_URL`, so if GitHub did make that change, regardless of what we do it'd break provenance for everyone using any currently released version of NPM.

So all this check does is make buggy but still widely used NPM versions between 9.5.0 and 9.6.0 behave the same as the fixed version NPM 9.6.1; it'll have no effect at all on any NPM versions beyond 9.6.0, and doesn't add any brittleness to the affected versions that wasn't already there.

src/release.js
AlanSl: Maybe that's better done as a small separate PR since it affects code out of scope of this feature.

README.md
```diff
@@ -217,6 +218,20 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
+ - Ensure your `package.json` is complete and correct. NPM will cancel the release with specific errors if it finds a problem. Requirements may change in future NPM releases but include things like a `"repository"` field with `"url"` property matching the format `"git+https://github.com/user/repo"`.
```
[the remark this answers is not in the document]
AlanSl: Sadly they're not documented (nor is it straightforward to see requirements in the source).

I've reworked this section in the ["Rework readme advice" commit](https://github.com/nearform-actions/optic-release-automation-action/pull/294/commits/11cc1d5499e5f719bee22d4e6120d25a7e65c933).

README.md
```diff
@@ -217,6 +218,20 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
+
+If `provenance: true` is in your YML inputs and any condition isn't met, the release will cancel with an error.
```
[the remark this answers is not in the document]
AlanSl: 👍 I've reworked this section in the ["Rework readme advice" commit](https://github.com/nearform-actions/optic-release-automation-action/pull/294/commits/11cc1d5499e5f719bee22d4e6120d25a7e65c933)

README.md
AlanSl: 👍 I've reworked this section in the ["Rework readme advice" commit](https://github.com/nearform-actions/optic-release-automation-action/pull/294/commits/11cc1d5499e5f719bee22d4e6120d25a7e65c933)

README.md
```diff
@@ -84,14 +84,20 @@ jobs:
...
       - uses: nearform-actions/optic-release-automation-action@v4
         with:
-          npm-token: ${{ secrets.NPM_TOKEN }}
-          optic-token: ${{ secrets.OPTIC_TOKEN }}
           commit-message: ${{ github.event.inputs.commit-message }}
           semver: ${{ github.event.inputs.semver }}
           npm-tag: ${{ github.event.inputs.tag }}
+          # optional: set this secret in your repo config for publishing to NPM
+          npm-token: ${{ secrets.NPM_TOKEN }}
+          # optional: set this secret in your repo config for 2FA with Optic
```
AlanSl, line 97: These lines were described as optional previously in the bullets below:

> _(Optional)_ You can define Npm and Optic tokens in GitHub secrets for each user that will receive the OTP. This is required only in case you want to publish to Npm.

...so now there's more optional config, I've moved all the optional lines together with comments

README.md
```diff
@@ -217,6 +225,43 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
+    steps:
+      - uses: nearform-actions/optic-release-automation-action@v4
+        with:
+          npm-token: ${{ secrets.NPM_TOKEN }}
+          optic-token: ${{ secrets.OPTIC_TOKEN }}
+          commit-message: ${{ github.event.inputs.commit-message }}
+          semver: ${{ github.event.inputs.semver }}
+          npm-tag: ${{ github.event.inputs.tag }}
+          # add this to activate the action's provenance feature
+          provenance: true
```
simoneb, line 261: wondering if there's a way to highlight certain lines using github code highlighting

simoneb (approved):
I'm approving with an outstanding question, see inline comments.

README.md
AlanSl: Seems not, sadly:

 - There's [an open feature request for it](https://github.com/orgs/community/discussions/42489): 
 - There's a [hacky workaround](https://stackoverflow.com/a/34538398) but it clashes with syntax highlighting and strips styling, only supports hard-to-see bolding of text in non-syntax-highlighted code blocks.

README.md
```diff
@@ -217,6 +225,43 @@ Generally, if you want to release a prerelease of a repository, and it is an NPM
...
 Please note that in case of a prerelease the `sync-semver-tags` input will be treated as `false`, even if it's set to `true`. This because we don't want to update the main version tags to the latest prerelease commit but only to the latest official release.
 
+## Provenance
+
+If `provenance: true` is added to your `release.yml`'s **inputs**, NPM will [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements).
+
+NPM has some internal [requirements](https://docs.npmjs.com/generating-provenance-statements#prerequisites) for generating provenance. Unfortunately as of May 2023, not all are documented by NPM; some key requirements are:
+
+- `id-token: write` must be added to your `release.yml`'s **permissions**
+- NPM must on version 9.5.0 or greater (this will be met if our recommended `runs-on: ubuntu-latest` is used)
```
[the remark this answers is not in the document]
AlanSl: I'd say no because it adds maintenance burden on our side (we'd need to choose when to update now and in future) and seems to go against least-surprise: right now, there's a very clear division of responsibility, where the user-side yml sets up the environment and runner, then our script runs inside their environment.

I'd thought about this, there's a few ways we could do it if we wanted to, the least obtrusive I could think of was adding something like `"engines": { "npm": ">=9.5.0" }` to package.json, which is better than specifying a node version because it allows users to be more up to date (e.g. they can jump to Node 20 before we do), and it doesn't impose on node version when all we have any reason to care about is NPM version.

But even that felt a bit heavy-handed for an opt-in feature. If a user wants an old NPM version (e.g. IIRC some projects stick to 8.5 to avoid side effects of changes to dependency resolution), and they don't care about provenance, there's really nothing wrong with that.

AlanSl:
@simoneb I've responded to the questions and pushed a couple of typo fixes I saw. If you're happy for it to be merged, looks like you need to do the final squash-merge: 

image

simoneb:
Yep thanks I agree with your conclusions, thanks!

simoneb (approved):

Issue #295, opened 2023-05-03
Prevent leaking secrets by stripping `INPUT_*` env vars when running @actions/exec 


AlanSl:
This is something I noticed while working on provenance.

By default, `@actions/exec`'s `exec` method copies all env vars to the executed child process, unless `env` is passed as an option, in which case it only includes those.

So by default, it passes all `INPUT_` env vars, which come from the `with` properties set in the user's `release.yml`. This is undesirable for two reasons:

 - They might contain repo secrets, so could be a vector where something like a github token can leak to an externally-controlled process (pretty sure we trust everything we `exec` but it's not great for this to be the default).
 - They're almost certainly irrelevant to the child process because they are custom inputs to this action

This is discussed by GitHub Actions team here - https://github.com/actions/toolkit/issues/309 - their conclusion is:

- They broadly agree that this isn't a good default but don't want to push a breaking change to 1.x
- They'll probably change this in 2.x but there's no roadmap or plan to actually do this
- For now they recommend stripping `INPUT_*` env vars manually

So it's probably a good idea for us to strip these vars on our side.

It's not burning-urgent because we currently only `exec` to `npm`, `git` and `curl` (and if we can't trust those we have bigger problems...) but it's something I think we should tighten up so we don't have to worry about third party services a future `exec` call might pass our inputs to..

PR #296, opened 2023-05-03
fix(issue 295): ensure inputs don't leak to child processes

AlanSl:
closes https://github.com/nearform-actions/optic-release-automation-action/issues/295

I split this out from the provenance work because it's a separate issue, posting this as an example implementation if we do think it's something worth doing. Not sure if it should have tests or not or if it's really possible to test?

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -79558,16 +79558,19 @@ const { exec } = __nccwpck_require__(1514)
  * @param {{cwd?: string}} options
  * @returns Promise<string>
  */
-async function execWithOutput(cmd, args, { cwd } = {}) {
+async function execWithOutput(
+  cmd,
+  args,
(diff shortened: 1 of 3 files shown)


simoneb (commented):
Is it feasible/valuable to have a test for this?

AlanSl:
Not sure. Ultimately it depends on an internal behaviour of GitHub Actions that ends in a child process. so I don't think we can test that? At least I don't think we can do any kind of full integration test here.

Maybe we can and should have a basic unit test that stubs `@actions/exec`, sets some env vars similar to how I did in the provenance tests, some of which are INPUT_ and some aren't, calls `execWithOutput`, and ensures that the exec is called with the non-input properties present and the input properties missing?

simoneb:
I'm sure you noticed by CI is red @AlanSl

simoneb (approved):

Issue #301, opened 2023-05-16
Can't release a newly created package with provenance


simoneb:
It looks like newly created packages cannot be published with this action and provenance enabled. 

See the output of this build which explains why: https://github.com/nearform/fastify-jwt-jwks/actions/runs/4993014295/jobs/8941522625

AlanSl:
🤔 Optic already does some logic to check on whether the package is already on NPM, in `allowNpmPublish`, as part of the check for whether this specific version already exists. We could tweak it to ignore the provenance flag in the `!packageName` condition (maybe logging something in the output), so people can just set `provenance: true` once and not need to change their release config for their first release?

Or maybe this is just something to document?

Though I feel like it'd be nicer if our default config could be something where you can just set it once when first setting up a project and it'll just work and (usually) not need to go in and change.

AlanSl:
Thinking about this again after the discussion this morning - is the underlying issue that on NPM "By default, scoped packages are published with private visibility" (something I often forget, and I'm sure others do too)?

I think we can infer that, if someone has set provenance as true, they must want the package to be published publicly, therefore, if provenance is true but we saw that the package isn't already on NPM, we should simply add `--access public` to the flags when the script runs `npm publish`?

simoneb:
@AlanSl that's the thinking that I was getting to too. Before doing this change though, I would:

1. make extra sure by reading the docs that provenance really is only for public packages, because I'm really not sure about this. If it is now, can we be certain that it will be in the future? I mean setting a package's visibility as public on behalf of the user is a BIG implication and if we do it, we must be extremely confident that that's the right thing to do
2. in case we go with that option, we should make it very visible in the docs of the action that this is what we'll do if provenance is enabled

In case we decide that the above is not a sensible decision, we may consider adding an additional input to the action, called indeed `access`, and we let users know that this needs to be set to `public` if they want to publish a public package with provenance enabled. Not great from a user's perspective to have to do 2 things but I'd be conservative if we're not sure.

AlanSl:
> setting a package's visibility as public on behalf of the user is a BIG implication and if we do it, we must be extremely confident that that's the right thing to do

100% agree.

How does Optic handles the first publish of a scoped package that should be public (without/before provenance)? 

IIRC when I used it with my package the first time (using Optic as it was without provenance support), it failed and I couldn't find any way to set access to public within Optic, so to save time since it wasn't my focus I just did `npm publish --access public` from the command line, then used Optic after.

[...]

PR #302, opened 2023-05-18
fix: add public inpuput to allow provenance

claudia-codes:
Solves #301 

To un-draft this: 
- [ ] solution must be validated
- [ ] tests must pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,9 @@ jobs:
           optic-token: ${{ secrets.OPTIC_TOKEN }}
           # optional: NPM will generate provenance statement, or abort release if it can't 
           provenance: true
+           # optional: requires to be public in order to enable provenance
+           # caution: this will make your npm package public
+          access: "public"
 ```
 
 The above workflow (when manually triggered) will:
@@ -117,7 +120,7 @@ When you merge this PR:
 - Upon successful retrieval of the OTP, it will publish the package to Npm.
 - Create a Github release with change logs (You can customize release notes using [release.yml](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration))
 - Leave a comment on each issues that are linked to the pull reqeuests of this release. This feature can be turned off by the `notify-on-the-issue` flag.
-- _(Optional)_ If `provenance: true` was set, NPM will add a [Provenance](#provenance) notice to the package's public NPM page.
+- _(Optional)_ If `provenance: true` and npm package access is public, NPM will add a [Provenance](#provenance) notice to the package's public NPM page. *New packages will disable provenance.*
 
 When you close the PR without merging it: nothing will happen.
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -284,7 +291,9 @@ jobs:
...
-| `provenance`| No    | Set as true to have NPM [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements.<br /> (_Default: `false`_)                                                                  |
+| `provenance`| No    | Set as true to have NPM [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements.<br /> (_Default: `false`_)      
+                                                                                                                   |
+| `provenance`| No    | Set as `public` to allow [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements.<br /> (_Default: `private`_)                                                                  |
```
AlanSl, line 296:
```suggestion
| `access`| No    | Set as `public` to allow [generate a provenance statement](https://docs.npmjs.com/generating-provenance-statements). See [Provenance section above](#provenance) for requirements.<br /> (_Default: `private`_)                                                                  |
```

simoneb (commented):
so far the implementation looks good if we are in agreement that we can't just enable public access implicitly, which I'm still not 100% sure about and I'd have to check. in any case, so far so good

src/utils/publishToNpm.js
```diff
@@ -41,24 +41,47 @@ async function allowNpmPublish(version) {
...
     'set',
     `//registry.npmjs.org/:_authToken=${npmToken}`,
   ])
 
+  let packageName = null
+  try {
+    const packageInfo = await execWithOutput('npm', ['view', '--json'])
+    packageName = packageInfo ? JSON.parse(packageInfo).name : null
+  } catch (error) {
+    if (!error?.message?.match(/code E404/)) {
```
AlanSl, line 76: Nit, but I wonder if we can avoid duplicating this, e.g. have the function that calls it already return a status number. or pull it out into its own utility that passes the "is published" state into `allowNpmPublish`?

AlanSl (commented):
I think the approach makes sense, though one question in addition to the other comments: are we expecting to add a feature where changing the `access` setting actually changes the access on NPM?

It feels like having this option creates that expectation, so if that's not part of the plan it needs to be clearly documented I think.

README.md
AlanSl: That's what I meant, "_Default: `private`_" seems to imply that if it is not defined, it'll make a package private.

simoneb:
I'm not sure we should be going ahead with this change because the requirements pertaining provenance and non-public packages are not clear.

If we want to create a new "access" input, then we should implement it fully, meaning that it should always be provided to the npm publish command, regardless of whether we use provenance or not.

I think we need to experiment more with this. E.g. does having a "private" property in package.json affect this behavior and would it be enough to publish with provenance without having to change this action? I don't know.

PR #305, opened 2023-05-23
feat: add `access`, fix provenance if new & unscoped

AlanSl:
closes https://github.com/nearform-actions/optic-release-automation-action/issues/301

This follows on from https://github.com/nearform-actions/optic-release-automation-action/pull/302 (starting a new branch / PR as we're both on personal forks) and does two related things:

- [x] Allows users to manually set the access control of their package in the input options, like `access: public` or `access: restricted`
- [x] Fixes https://github.com/nearform-actions/optic-release-automation-action/issues/301 and allows never-published unscoped packages to be published with provenance in their first release, by inserting `--access: public` if and only if:
   - the user hasn't specified access control in any way, and
   - the package is unscoped (therefore, inherently public and cannot be private / restricted), and
   - the package hasn't been published before (so NPM's provenance feature would fail to see it as public unless we do this)

This would be a **minor version bump** on the action: the access control is a new feature, there's no breaking change or requirement to use the new feature because the provenance fix only has an affect when the package would have been public anyway.

Here's an example of an unscoped package published for the first time with provenance using this, without explicitly (redundantly) specifying the package to be public: https://www.npmjs.com/package/alansl-new-release-tester-2/v/0.1.0

---

@simoneb looks like I can't add you via "assign reviews" so I'll ping you here

### DIFF
--- a/src/release.js
+++ b/src/release.js
@@ -3,7 +3,7 @@
 const core = require('@actions/core')
 const semver = require('semver')
 
-const { PR_TITLE_PREFIX } = require('./const')
+const { ACCESS_OPTIONS, PR_TITLE_PREFIX } = require('./const')
 const { callApi } = require('./utils/callApi')
 const { tagVersionInGit } = require('./utils/tagVersion')
 const { revertCommit } = require('./utils/revertCommit')
@@ -103,10 +103,9 @@ module.exports = async function ({ github, context, inputs }) {
     const access = inputs['access']
 
     // Can't limit custom action inputs to fixed options like "choice" type in a manual action
-    const validAccessOptions = ['public', 'restricted']
-    if (access && !validAccessOptions.includes(access)) {
+    if (access && !ACCESS_OPTIONS.includes(access)) {
       core.setFailed(
-        `Invalid "access" option provided ("${access}"), should be one of "${validAccessOptions.join(
+        `Invalid "access" option provided ("${access}"), should be one of "${ACCESS_OPTIONS.join(
           '", "'
         )}"`
       )
--- a/src/release.js
+++ b/src/release.js
@@ -113,7 +113,7 @@ module.exports = async function ({ github, context, inputs }) {
       return
     }
 
-    let publishOptions = {
+    const publishOptions = {
       npmToken,
       opticToken,
       opticUrl,
@@ -127,9 +127,10 @@ module.exports = async function ({ github, context, inputs }) {
       // Fail fast with meaningful error if user wants provenance but their setup won't deliver,
       // and apply any necessary options tweaks.
       const npmVersion = await getNpmVersion()
-      publishOptions = await ensureProvenanceViability(
-        npmVersion,
-        publishOptions
+
+      Object.assign(
+        publishOptions,
+        await ensureProvenanceViability(npmVersion, publishOptions)
       )
     }
 
--- a/src/release.js
+++ b/src/release.js
@@ -102,7 +102,6 @@ module.exports = async function ({ github, context, inputs }) {
     const provenance = /true/i.test(inputs['provenance'])
     const access = inputs['access']
 
-    // Can't limit custom action inputs to fixed options like "choice" type in a manual action
     if (access && !ACCESS_OPTIONS.includes(access)) {
       core.setFailed(
         `Invalid "access" option provided ("${access}"), should be one of "${ACCESS_OPTIONS.join(
--- a/src/release.js
+++ b/src/release.js
@@ -11,10 +11,7 @@ const { publishToNpm } = require('./utils/publishToNpm')
 const { notifyIssues } = require('./utils/notifyIssues')
 const { logError, logInfo, logWarning } = require('./log')
 const { execWithOutput } = require('./utils/execWithOutput')
-const {
-  ensureProvenanceViability,
-  getNpmVersion,
-} = require('./utils/provenance')
+const { getProvenanceOptions, getNpmVersion } = require('./utils/provenance')
 
 module.exports = async function ({ github, context, inputs }) {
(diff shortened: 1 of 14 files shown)


test/publishToNpm.test.js
```diff
@@ -90,15 +100,18 @@ tap.test(
       'pack',
       '--dry-run',
     ])
+    t.pass('npm pack called')
```
AlanSl, line 103: I added a lot of these as without at least one test, Tap's output doesn't say which test in a file failed

src/release.js
```diff
@@ -100,22 +100,41 @@ module.exports = async function ({ github, context, inputs }) {
     const opticToken = inputs['optic-token']
     const npmToken = inputs['npm-token']
     const provenance = /true/i.test(inputs['provenance'])
+    const access = inputs['access']
+
+    // Can't limit custom action inputs to fixed options like "choice" type in a manual action
+    const validAccessOptions = ['public', 'restricted']
```
simoneb, line 106: consider moving these outside of the function as it's static data

src/release.js
```diff
@@ -100,22 +100,41 @@ module.exports = async function ({ github, context, inputs }) {
...
+    if (access && !validAccessOptions.includes(access)) {
+      core.setFailed(
+        `Invalid "access" option provided ("${access}"), should be one of "${validAccessOptions.join(
+          '", "'
+        )}"`
+      )
+      return
+    }
+
+    let publishOptions = {
```
simoneb, line 116: prefer const when you don't need mutable state

src/release.js
```diff
@@ -100,22 +100,41 @@ module.exports = async function ({ github, context, inputs }) {
...
-    // Fail fast with meaningful error if user wants provenance but their setup won't deliver
     if (provenance) {
+      // Fail fast with meaningful error if user wants provenance but their setup won't deliver,
+      // and apply any necessary options tweaks.
       const npmVersion = await getNpmVersion()
-      checkProvenanceViability(npmVersion)
+      publishOptions = await ensureProvenanceViability(
+        npmVersion,
+        publishOptions
+      )
```
simoneb, line 133: can we mutate the options provided to this function instead of returning them and having to overwrite the publishOptions variable?

src/utils/provenance.js
```diff
@@ -40,18 +41,53 @@ function checkPermissions(npmVersion) {
...
   checkIsSupported(npmVersion)
   checkPermissions(npmVersion)
-  // There are various other provenance requirements, such as specific package.json properties, but these
-  // may change in future NPM versions, and do fail with meaningful errors, so we let NPM handle those.
+
+  const value = {
+    ...publishOptions,
+    ...(await getAccessAdjustment(publishOptions)),
+  }
+  return value
```
simoneb, line 90: in order to avoid mutating the state of the called function, just use Object.assign

simoneb (approved):
A few minor code changes if you don't mind, approving to avoid blocking. Overall LGTM, thanks for all the work and testing

AlanSl:
Comments all make sense, I've simplified the options overrides and moved that fixed array to const.js with other constants

simoneb (commented):
A few more comments. Would you mind testing the weird tap behavior in isolation to make sure we're not looking at a red herring? I would be surprised if an exception being thrown in a test (and not caused by a failing tap assertion) resulted in tap not printing the test where it occurred.

test/packageInfo.test.js
```diff
@@ -0,0 +1,117 @@
...
+    t.match(packageInfo, null)
+  }
+)
+
+tap.test('getPublishedInfo throws if it hits a non-404 error', async t => {
+  const mocks = setupPublished({
+    error: new Error('code E418 - unexpected teapot'),
+  })
+
+  t.rejects(mocks.getPublishedInfo, /teapot/)
```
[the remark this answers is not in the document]
AlanSl: Makes sense.

While mopping up my own instances of this, I've found a few pre-existing examples where I think tests aren't behaving as intended, e.g. there's several with a pattern like this in the `execWithOutput` tests from a few months ago:

```js
    t.resolves(execWithOutputModule.execWithOutput('ls', ['-al']), output)
    execStub.calledWith('ls', ['-al'])
```

I think these are intended to be like this:

```js
    await t.resolves(execWithOutputModule.execWithOutput('ls', ['-al']), output)
    t.ok(execStub.calledWith('ls', ['-al']))
```

...because as previously written, the `execStub.calledWith` will fire before the async exec actually happens, and just returns `false` into the void with no effect? I might fix these in a small separate PR.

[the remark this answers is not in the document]
simoneb: yes please, any async tap assertions need to be awaited for them to have any effect

src/utils/publishToNpm.js
```diff
@@ -1,21 +1,15 @@
...
-  } catch (error) {
-    if (!error?.message?.match(/code E404/)) {
-      throw error
-    }
-  }
 
+  const packageInfo = await getPublishedInfo()
+  const packageName = packageInfo?.name
   // Package has not been published before
   if (!packageName) {
```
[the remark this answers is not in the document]
AlanSl: 🤔 it's probably impossible in real life for`npm view --json` to return an object that had no `name` property, but if it did and we did this, further down we'd exec `npm --view undefined@x.y.x`.

Can simplify it to `if (!packageInfo?.name) return true`(and `${packageInfo.name}@${version}` below) I guess, that'd still be type safe for what the external service returns.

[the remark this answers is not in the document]
simoneb: fair enough

AlanSl:
@simoneb I think I responded to all the comments, ready for re-review / merge (I don't have access to merge it myself)

PR #308, opened 2023-05-25
chore: await async tests, assert exec calls

AlanSl:
This came up in review of https://github.com/nearform-actions/optic-release-automation-action/pull/305 (specifically [this comment thread](https://github.com/nearform-actions/optic-release-automation-action/pull/305#discussion_r1203712994))

> when you make async assertions (such as .rejects or .resolves), you need to either return or (preferably, in my opinion) await the result of the assertion, otherwise the assertion will be a noop

- [x] As discussed, there were a lot of pre-existing non-await `t.resolves` and `t.rejects` in tests, so this tidies them up.
- [x] There were also some `stub.calledWith()` calls that look like they were supposed to be test assertions, but didn't actually assert anything, they just returned `true` or `false` into the void. I've replaced these with actual assertions.

### DIFF
--- a/test/archiver.test.js
+++ b/test/archiver.test.js
@@ -36,13 +36,13 @@ tap.test('throws an error if path not found', async t => {
     },
   })
 
-  t.rejects(archiverModule.archiveItem('path', 'out.zip'))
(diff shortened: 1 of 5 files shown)


simoneb (approved):
thanks for this

PR #310, opened 2023-05-31
feat: send package info when requesting otp

guilhermelimak:
This PR adds support for showing package name and version when receiving an OTP notification from optic. In this project the only change is that now the package name and version are sent as payload on the OTP request.

This PR needs to be merged after the [api](https://github.com/nearform/optic/pull/323) one since we changed the request method from GET to POST there.

References nearform/optic-expo#1186.


src/utils/publishToNpm.js
```diff
@@ -62,9 +62,17 @@ async function publishToNpm({
 
   if (await allowNpmPublish(version)) {
     await execWithOutput('npm', ['pack', '--dry-run'])
+
     if (opticToken) {
+      const packageInfo = await getPublishedInfo()
```
simoneb, line 67: can you make extra sure that:

- we don't call this function unnecessarily, e.g. if the information is already available to us
- in case we do need to retrieve this info, it is safe to do so in this place

guilhermelimak: you're correct, I think we can just get it from the local package.json file here. Just updated the code.

simoneb (approved):

PR #339, opened 2023-07-24
Add error reporting when release branch name conflicts with existing branch(es)

### LINKED ISSUE #329: Send appropriate error message when attempting to create a release branch that already exists.
Current behaviour: When the action attempts to create a release branch, but there is already a branch of the same name, the action fails. In this case it should give a specific error message detailing what happened, and how to fix it by deleting the branch with the same name.

This can happen if the release flow fails unexpectedly and leaves behind a lingering branch.

ollyjohn:
Closes #329 

### Manual testing carried out:
* [x] Create a release branch & add custom workflow that allows you to enter the version in plain text, specify a version for a new release whose branch already exists - should get rejected with a sensible message
* [x] Specify a version for a new release that ***doesn't*** already have a branch - should run the script to completion

Test runs can be found [here](https://github.com/nearform/ojj-optic-clash-test/actions)

### DIFF
--- a/src/openPr.js
+++ b/src/openPr.js
@@ -8,7 +8,7 @@ const core = require('@actions/core')
 const { PR_TITLE_PREFIX } = require('./const')
 const { callApi } = require('./utils/callApi')
 const transformCommitMessage = require('./utils/commitMessage')
-const { logError, logInfo, logWarning } = require('./log')
+const { logInfo, logWarning } = require('./log')
 const { attach } = require('./utils/artifact')
 const { getPRBody } = require('./utils/releaseNotes')
 const { execWithOutput } = require('./utils/execWithOutput')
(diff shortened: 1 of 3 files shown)


src/openPr.js
```diff
@@ -95,7 +95,14 @@ module.exports = async function ({ context, inputs, packageVersion }) {
...
 
   const messageTemplate = inputs['commit-message']
-  await execWithOutput('git', ['checkout', '-b', branchName])
+  try {
+    await execWithOutput('git', ['checkout', '-b', branchName])
+  } catch (error) {
+    core.setFailed(`${error.message}`)
+    const message = `${error.message}.  Please delete any branch with this name that may already exist and try again.`
+    logError(message)
+    throw new Error(message)
```
simoneb, line 104: I think this error handling is redundant. throwing an error and calling core.setFailed both interrupt the execution of the action. I suppose they also both log a message, meaning that explicit logging may not be necessary.

trying it out locally is not easy, but you can set up a dummy repo, point it at this branch and try to reproduce the error, and see how it fails

ollyjohn: You're right, it also turned out that the `checkout -b` command wasn't actually erroring as the branch didn't exist locally so I've revised my approach to check the remote before attempting to do anything else, which seems to work nicely

test/bump.test.js
```diff
@@ -115,6 +127,29 @@ tap.test('it triggers an error when the packageVersion is missing', async t => {
...
+  try {
+    await openPr(actionData)
+    t.fail('Should have thrown an error')
+  } catch (error) {
+    sinon.assert.match(
+      error.message,
+      'Release branch release/v1.2.3 already exists on the remote.  Please either delete it and run again, or select a different version'
+    )
+  }
+})
```
simoneb, line 151: use t.rejects to test for promise rejection

test/bump.test.js
```diff
@@ -190,7 +231,7 @@ tap.test('should work with a custom version-prefix', async () => {
   sinon.assert.calledWithExactly(stubs.execWithOutputStub, 'git', [
     'commit',
     '-m',
-    `Release v${TEST_VERSION}`,
+    `Release ${TEST_VERSION}`,
```
simoneb, line 234: why would this assertion be different from what it was before?

ollyjohn: I think this test was passing as a false positive previously - on the `main` branch in the nearform-actions repo if you run it as the only (or first) test in the file it fails and, following the code through, it shouldn't have the `v` if you pass `inputs['version-prefix']` as `''`.

Having a look into why this is happening now, feels like maybe the mocks aren't getting restored/reset properly after each tests

simoneb: yes please let's spend some time trying to figure out what's going on here to avoid surprises

simoneb (approved):
LGTM

PR #413, opened 2024-01-25
fix: fixes OTP leak in case of error

vikasbhandari2:
### What does this PR do?

Since the OTP is a confidential information, it should not be visible in the logs. 

This PR fixes the leak and adds the code to redact the confidential arguments so it doesn't print on the Build Error Logs. The supporting tests have been added as well.


src/utils/execWithOutput.js
```diff
@@ -3,6 +3,28 @@
...
+          skipItem = false
+          
+          return false;
+      }
+
+      skipItem = CONFIDENTIAL_KEYWORDS_FOR_REDACTION.includes(arg?.toString().toLocaleUpperCase())
+
+      return !skipItem
+  })
+}
```
simoneb, line 27: it's not very easy to follow what this code is doing, plus there are not tests for this function outside of the edge cases. let's please write this in a more idiomatic way so that understanding what it does it easier for a reader

test/execWithOutput.test.js
```diff
@@ -111,3 +112,43 @@ tap.test('passes env vars excluding `INPUT_*` env vars', async t => {
   // Its value will vary by test runner, so just check it is present.
   t.hasProp(envInExec, 'NODE')
 })
+
+tap.test('Invalid arguments inputs should not fail', async t => {
```
simoneb, line 116: we need to test for normal cases here, not just edge cases

test/execWithOutput.test.js
```diff
@@ -111,3 +112,43 @@ tap.test('passes env vars excluding `INPUT_*` env vars', async t => {
...
+  const redactedUndefinedArray = removeConfidentialArguments(undefined);
+  const redactedNullArray = removeConfidentialArguments(null);
+
+  t.equal(Array.isArray(redactedBlankArray), true)
+  t.equal(Array.isArray(redactedUndefinedArray), true)
+  t.equal(Array.isArray(redactedNullArray), true)
+
+  t.equal(redactedBlankArray?.length, 0)
+  t.equal(redactedUndefinedArray?.length, 0)
+  t.equal(redactedNullArray?.length, 0)
```
simoneb, line 127: the conditional operator is unnecessary, you have already tested that it's an array

test/execWithOutput.test.js
```diff
@@ -111,3 +112,43 @@ tap.test('passes env vars excluding `INPUT_*` env vars', async t => {
...
+
+  const errorMessage1 = await t.rejects(execWithOutputModule.execWithOutput('ls', arrayWithOTP));
+  const errorMessage2 = await t.rejects(execWithOutputModule.execWithOutput('ls', arrayWithOTPInStart));
+  const errorMessage3 = await t.rejects(execWithOutputModule.execWithOutput('ls', arrayWithOTPAtEnd));
+  const errorMessage4 = await t.rejects(execWithOutputModule.execWithOutput('ls', args));
+  
+  t.ok(errorMessage1.message.indexOf('--otp') === -1 && errorMessage1.message.indexOf('publish') > -1)
+  t.ok(errorMessage2.message.indexOf('--otp') === -1 && errorMessage2.message.indexOf('publish') > -1)
+  t.ok(errorMessage3.message.indexOf('--otp') === -1 && errorMessage3.message.indexOf('publish') > -1)
+  t.ok(errorMessage4.message.indexOf('publish') > -1)
```
simoneb, line 152: ideally test using the built in assertions, which will give you meaningful error messages when they fail. if you test way, the error message won't tell you anything about the nature of the failure, as it will only show you true or false

PR #438, opened 2024-05-15
Conventional changelog

0xor1:
https://github.com/nearform/hub-draft-issues/issues/361


package.json
```diff
@@ -21,7 +21,7 @@
     "@actions/exec": "^1.1.1",
     "@actions/github": "^6.0.0",
     "adm-zip": "^0.5.12",
-    "conventional-changelog-monorepo": "github:nearform/conventional-changelog",
+    "conventional-changelog-monorepo": "github:conventional-changelog/conventional-changelog",
```
simoneb, line 24: we need to reference the npm package now, not the github repo

src/index.js
```diff
@@ -51,10 +49,8 @@ async function getAutoBumpedVersion(baseTag) {
 
   logInfo(`Using ${tag} as base release tag for version bump`)
 
-  const { releaseType = 'patch' } = await conventionalRecommendedBumpAsync({
-    baseTag: tag,
-    config: conventionalCommitsConfig,
-  })
+  const bumper = new Bumper(process.cwd()).tag(baseTag)
+  const { releaseType = 'patch' } = await bumper.bump()
```
simoneb, line 53: this change very much depends on what the new API is. I would have imagined that the API had remained largely unchanged, but that doesn't seem to be the case if this requires this change

Issue #231, opened 2023-01-09
Use the conventional-changelog package for specifying base tag


sameer-coder:
A [PR](https://github.com/conventional-changelog/conventional-changelog/pull/954) is opened here to add a new feature to the conventional-recommended-bump package of conventional-changelog package. There hasn't been any response on the PR so far.

This issue is created to serve as reminder to check if the PR has been merged and revert the base tag changes in PR(https://github.com/nearform/optic-release-automation-action/pull/202) and use the base tag option from the conventional-commits package

simoneb:
@sameer-coder isn't this supposed to be closed by the PR that was merged #202 ?

simoneb:
This is the issue we reported and that should have eventually led to a new version being published https://github.com/conventional-changelog/conventional-changelog/issues/1224

PR #449, opened 2024-06-12
feat: migrate to esm to use official conventional-recommended-bump

G100g:
Drop `conventional-changelog` fork in favor to the official `conventional-recommended-bump` package #231 

`conventional-recommended-bump` is published as ESM so the PR contains also the migration of the entire action from CMJ to ESM


G100g:
Too many conflicts to update. I don't think I have time to update the PR.
But I can reopen it if you think all is not yet lost... :D

PR #492, opened 2025-01-08
chore: alternative otp provisioning pseudo code

simoneb:


src/utils/publishToNpm.js
```diff
@@ -63,18 +63,57 @@ async function publishToNpm({
...
+      } else if (ngrokApiKey) {
+        // 1. run a fastify app which exposes a simple web UI with a text input to get the OTP
+        // 2. expose the app via ngrok
+        // 3. print the url to the app on the workflow execution output
+        // 4. wait for some time or a timeout
+        // 5. if we get the token from the user, continue with the execution, otherwise timeout
+        const app = fastify()
+
+        app.get('/', () => {
+          // serve static page with text input which, when submitted, POSTs the OTP to /otp url
```
simoneb, line 92: @nickzafiropulos @msgadi can you please work together to create a simple template for this page which is on-brand?

simoneb: @msgadi you could use fastify/static, if you wish, or maybe even keep it simple and just build a simple html page with css and js inline

src/utils/publishToNpm.js
```diff
@@ -63,18 +63,57 @@ async function publishToNpm({
...
+        app.post('/otp', async req => {
+          const otp = req.body.otp
+
+          otpPromiseResolve(otp)
+        })
+
+        app.listen({ host: ..., port: ...})
+
+        // expose the fastify app via nkgrok and get the ngrok app url, see https://github.com/ngrok/ngrok-javascript
+        const ngrokUrl = ...
```
simoneb, line 107: we said that it's probably easier to use localtunnel instead of ngrok, but the api is very similar anyway

nickzafiropulos:
@msgadi @simoneb 
I have suggested a design here - https://www.figma.com/design/SZygq93AtTJQHx1ajTmdls/Optic-Expo---UI-Refresh?node-id=35-3860

It's inspired by https://mappy.nearform.com/ - so some of the basic css (background, form container, elevation shadows) could probably be taken from that to make it easier?

nickzafiropulos:
> Thanks @nickzafiropulos , can you please enrich that page with a couple more bits of information? The user is prompted to provide a OTP for publishing a certain version of a certain package, so let's include that read-only information in the layout of the page:
> 
> * package name
> * package version

That's done - have a look again.. 
How long are the package names likely to be? If they're likely to be longer than that space, I'll drop that and the version number onto new lines and alter the styling a bit

simoneb:
yep that's good, they're usually fairly short as they're npm package names

src/utils/publishToNpm.js
```diff
@@ -63,18 +63,57 @@ async function publishToNpm({
...
+          JSON.stringify({ packageInfo: { version, name: packageInfo?.name } }),
+          '-H',
+          'Content-Type: application/json',
+          '-X',
+          'POST',
+          `${opticUrl}${opticToken}`,
+        ])
+      } else if (ngrokApiKey) {
+        // 1. run a fastify app which exposes a simple web UI with a text input to get the OTP
+        // 2. expose the app via ngrok
```
simoneb, line 85: we're having an issue here whereby we can't bundle ngrok, so we need to find a way to execute ngrok at runtime, rather than build time

simoneb: see the changes to the action.yml file for details

simoneb:
closing as this is being worked on already

PR #526, opened 2025-03-31
fix(517): handle optic and ngrok promise race

### LINKED ISSUE #517: Workflow execution hangs due to fastify server not being closed
We have noticed that when the publishing happens, we are waiting for the fastify server used to the ngrok-based publishing to close (we have a timeout in there) even though the Optic-app based publishing has already happened.

You can see an example here: https://github.com/nearform/netsuite-restlet-api/actions/runs/13652672786/job/38164615158

Image

Even though the publishing had alrady happened via the Optic mobile app, the workflow was still waiting on the ngrok callback before timing out.

We should build this in a way that whatever publishing comes first, the process continues and stops waiting for the other.

As part of this, let's also check if we're properly skipping ngrok installation unless a token is provided, because I'm under the impression that executions take a long time regardless (but I'm not 100% sure)

confraria:
fixes #517 
- adds an abort signal to reject ngrok promise when any otp promise settles which will close the fastify server that keeps the action running more time than needed and also supresses the timeout error log message

### DIFF
--- a/src/utils/ngrokOtpVerification.js
+++ b/src/utils/ngrokOtpVerification.js
@@ -5,11 +5,10 @@ const fastify = require('fastify')
 const { logInfo, logError } = require('../log')
 const getNgrok = require('./getNgrok')
 
-const otpHtml = fs.readFileSync(__dirname + '/assets/otp.html', 'utf8')
 const otpVerificationTimeout = 300000 // 5 minutes;
-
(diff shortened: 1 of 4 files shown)


simoneb (approved):
amazing 🚀

confraria:
> ha, we need full coverage please :)

pushed ✅

simoneb (approved):

